PR #2, opened 2016-03-01
Initial commit for security related config rules.

anthroprose:
Initial commit of security related config rules including language breakouts.


chrgut:
I left some feedback in the acorley-development branch

PR #6, opened 2016-03-07
Add rule that check vpc flow logs enabled

okochang:
This rules check that VPC Flow Logs is enabled at specified VPC

### DIFF
--- a/python/vpc_flow_logs_enabled.py
+++ b/python/vpc_flow_logs_enabled.py
@@ -5,11 +5,11 @@
 import boto3
 import json
 
-def evaluate_compliance(config_item, is_flow_logs_enabled):
+def evaluate_compliance(config_item, vpc_id):
     if (config_item['resourceType'] != 'AWS::EC2::VPC'):
         return 'NOT_APPLICABLE'
 
-    elif is_flow_logs_enabled:
+    elif is_flow_logs_enabled(vpc_id):
         return 'COMPLIANT'
     else:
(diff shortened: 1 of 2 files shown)


pprahlad:
r

python/vpc_flow_logs_enabled.py
```diff
@@ -0,0 +1,50 @@
...
+    if len(response[u'FlowLogs']) != 0: return True
+
+def lambda_handler(event, context):
+    invoking_event = json.loads(event['invokingEvent'])
+
+    compliance_value = 'NOT_APPLICABLE'
+
+    vpc_id = invoking_event['configurationItem']['resourceId']
+    compliance_value = evaluate_compliance(
+        invoking_event['configurationItem'], is_flow_logs_enabled(vpc_id))
```
aaronkhoo: The way this is written, is_flow_logs_enabled will be called for all configurationItems passed to this lambda_handler, since it is evaluated before the outer function (evaluate_compliance) is called. Since  is_flow_logs_enabled does a ec2.describe_flow_logs service call each time, this could be quite inefficient. Suggest refactoring code to explicitly test for applicability first before calling is_flow_logs_enabled(vpc_id)

okochang: Thank you for proposal !
I will fix my code soon.

aaronkhoo:
Looks good. Merging.

PR #8, opened 2016-03-10
AWS CIS Benchmark 1.9

thigley986:
AWS CIS Benchmark 1.9
1.9 Ensure IAM password policy requires minimum length of 14 or greater (Scored)

This file made available under CC0 1.0 Universal (https://creativecommons.org/publicdomain/zero/1.0/legalcode)


aaronkhoo:
By convention, our file names should take the form of description-ruletype, so this should be something  like "password_policy_min_password_length-periodic.py". This helps users figure out the intent of the rule and how they should configure it (periodic vs triggered). Could you please update the file name appropriately?

thigley986:
I have updated to:
password_policy_min_password_length_CIS_benchmark_1_9-periodic

Is this in proper conformance with convention?

aaronkhoo:
Looks good. Merging.

Issue #7, opened 2016-03-10
Documentation Updates


anthroprose:
Within the code and also Rules.MD, add comments that indicate
    * what type of rule this (periodic/trigger) is and
    * Recommended scope (resource type). Can be high level like IAM but needed to avoid log spamming and complicating troubleshooting.
    * IAM policy needed

PR #9, opened 2016-03-14
Documentation Updates

anthroprose:
Adding documentation to each lambda script.

Adding new details to RULES.md

https://github.com/awslabs/aws-config-rules/issues/7

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -5,13 +5,15 @@ Description: Checks to see if there is a password policy section enabled in IAM.
 
 	node/iam_password_policy_enabled-periodic.js
 
-Required Paramters:```None```
+Trigger Type: ```Periodic```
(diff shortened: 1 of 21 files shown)


RULES.md
```diff
@@ -71,14 +81,17 @@ Example Value: ```dedicated```
 Description: Checks that CloudTrail is enabled in all regions. Use this rule only in your home region
```
pprahlad: Not your change, but this description seems incorrect for the rule specified. This should read something like
"Description: Checks that the tenancy of the EC2 Instances in this account are of type specified in DesiredTenancy, and reports complaince on the EC2 instance

pprahlad:
Looks good. One correction suggested

PR #18, opened 2016-03-29
Correct MaxPasswordLength to MaxPasswordage

nickdirienzo:
For `node/iam_password_maximum_age-periodic.js`, the parameter key seems to be incorrect. This should resolve the documentation inconsistencies. 


aaronkhoo:
Good catch. Thanks for fixing.

PR #22, opened 2016-04-14
Add python rules

gguillemas:
I wanted to contribute eight rules (one of which will only find applicable resources once AWS Config supports S3 buckets) that I developed for my company, Schibsted. I confirm these files are made available under CC0 1.0 Universal (https://creativecommons.org/publicdomain/zero/1.0/legalcode).


python/s3-exposed-bucket.py
```diff
@@ -0,0 +1,84 @@
+#
+# This file made available under CC0 1.0 Universal (https://creativecommons.org/publicdomain/zero/1.0/legalcode)
+#
+# Ensure that no buckets are globally accessible.
+# Description: Checks that all buckets restrict access from the global groups.
+#
+# Trigger Type: Change Triggered
+# Scope of Changes: S3:Bucket
+#
+# NOTE: This rule will only find applicable resources once S3 resources are supported by AWS Config.
```
aaronkhoo: Thanks for the enthusiasm in creating a rule for S3 buckets. However, since S3 resources are not currently supported by AWS Config, this could cause user confusion. Could you please remove this until AWS Config supports S3 resources officially?

python/ec2-exposed-group.py
```diff
@@ -0,0 +1,103 @@
...
+def evaluate_compliance(configuration_item, rule_parameters):
+    if configuration_item["resourceType"] not in APPLICABLE_RESOURCES:
+        return {
+            "compliance_type": "NOT_APPLICABLE",
+            "annotation": "The rule doesn't apply to resources of type " +
+            configuration_item["resourceType"] + "."
+        }
+
+    ec2 = boto3.resource("ec2")
+    security_group = ec2.SecurityGroup(configuration_item["resourceId"])
```
aaronkhoo: The ipPermissions are actually available as part of the structure in invokingEvent. You can access them by doing:

configuration = configuration_item['configuration']
ipPerms = configuration['ipPermissions']

You can see the configuration_item for a resource on the AWS Config console by clicking on its timeline, and then clicking on the "View Details" link. Look for the "configuration" element, and the "ipPermissions" element should be below that.

python/ec2-exposed-instance.py
```diff
@@ -0,0 +1,110 @@
...
+    if configuration_item["resourceType"] not in APPLICABLE_RESOURCES:
+        return {
+            "compliance_type": "NOT_APPLICABLE",
+            "annotation": "The rule doesn't apply to resources of type " +
+            configuration_item["resourceType"] + "."
+        }
+
+    ec2 = boto3.resource("ec2")
+    instance = ec2.Instance(configuration_item["resourceId"])
+    security_groups = instance.security_groups
```
aaronkhoo: See comment above for security groups. In this case, the security groups for an instance are available as part of configuration_item['configuration']['securityGroups']. You do have to use ec.SecurityGroups to retrieve the ip_permissions as you do below, however, since the configuration doesn't contain that information.

aaronkhoo:
Thanks for submitting a great set of rules. We had a few comments in the code.

python/s3-exposed-bucket.py
gguillemas: Thank you very much for your comment. You are right, I agree that it could be confusing for some users, I have updated the PR without this rule. I'll save it for when S3 buckets become officially supported by AWS Config.

python/ec2-exposed-group.py
gguillemas: Thank you very much for your suggestion, it will be very useful in the future. I have updated both exposed rules to use the configuration attribute and also the MFA rule, which now doesn't have to rely on a hack to obtain the username of the resource.

gguillemas:
Thank you for your suggestions, I have updated the PR with the new changes.

python/ec2-exposed-group.py
aaronkhoo: Thanks for updating. We're trying to make sure this is clear to developers that the "configuration" element actually has a describe of the resource in the configuration_item already, so they can save a service call to retrieve this info. Appreciate you modifying the code to help make this clear.

aaronkhoo:
Looks good. Merging.

PR #23, opened 2016-04-15
Add security-group by tag python rule

aaronkhoo:
Add a new python rule that ensures all EC2 Instances that match a certain tag regex format also have a specific security group attached.

### DIFF
--- a/python/ec2_require_security_group_by_tag.py
+++ b/python/ec2_require_security_group_by_tag.py
@@ -1,0 +1,71 @@
+#
+# This file made available under CC0 1.0 Universal (https://creativecommons.org/publicdomain/zero/1.0/legalcode)
+#
+# Ensure all EC2 Instances that have a certain tag format also have a specific security group
+# Description: Checks that all EC2 instances that have a certain tag format also have a specific security group
+#
+# Trigger Type: Change Triggered
+# Scope of Changes: EC2:Instance
+# Required Parameters: name
+# Example Value: ^prod(us|eu|br)[lw]box[0-9]{3}$ (which will match 'produslbox001')
+# Required Parameters: secGroup
+# Example Value: MySecGroup
+# 
+
+import boto3
+import json
+import re
+
+def is_applicable(config_item, event):
+    status = config_item['configurationItemStatus']
+    event_left_scope = event['eventLeftScope']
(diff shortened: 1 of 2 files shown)


pprahlad:
Would recommend "TagName" as the rule parameter because name can be confused with the instance name

aaronkhoo:
@pprahlad Actually I am targeting the instance name specifically (which shows up as a tag with key value "name").

python/ec2_require_security_group_by_tag.py
```diff
@@ -0,0 +1,71 @@
...
+# 
+
+import boto3
+import json
+import re
+
+def is_applicable(config_item, event):
+    status = config_item['configurationItemStatus']
+    event_left_scope = event['eventLeftScope']
+    test = ((status in ['OK', 'ResourceDiscovered']) and
```
ryflood: variable name 'test' is a bit vague. Please consider an alternative, or simply return from this line

python/ec2_require_security_group_by_tag.py
```diff
@@ -0,0 +1,71 @@
...
+            event_left_scope == False)
+    return test
+
+def evaluate_compliance(config_item, rule_parameters):
+    # Initialize evaluation to 'not applicable', i.e. rule doesn't apply
+    evaluation = 'NOT_APPLICABLE'
+    if (config_item['resourceType'] == 'AWS::EC2::Instance'):
```
ryflood: this check may make more sense in the 'is_applicable' method

python/ec2_require_security_group_by_tag.py
```diff
@@ -0,0 +1,71 @@
...
+        configuration = config_item['configuration']
+        tags = configuration['tags']
+        foundTag = False
```
ryflood: is this variable required?

python/ec2_require_security_group_by_tag.py
```diff
@@ -0,0 +1,71 @@
...
+        # If the config item is for an EC2 instance, then iterate through the tags for that instance
+        for tag in tags:
+            # Check if this is the name tag, and that it matches the provided regex value
```
ryflood: [nit] - "Check if this is the 'Name' tag..." or "Check if this tag has key 'Name'..." may be more readable

PR #26, opened 2016-04-21
Changed readSnapshot  to retrieve s3 bucket region prior to retrieval

Sam-Martin:
Created constructor for readSnapshot to retrieve s3 bucket region to stop `PermanentRedirect: The bucket you are attempting to access must be addressed using the specified endpoint.` error when accessing config bucket from another region.  
Because the AWS Config bucket can be configured as a bucket in an external account it's possible that the home region for CloudTrail and the AWS Config bucket region may not match. Therefore I've added a constructor which finds the config bucket's S3 region in order to allow the getObject call to address the correct endpoint.  

I appreciate this PR is a little larger than my previous ones, but I welcome any feedback you have on alternative approaches!  

### DIFF
--- a/node/cloudtrail_enabled_all_regions-periodic.js
+++ b/node/cloudtrail_enabled_all_regions-periodic.js
@@ -78,7 +78,7 @@ readSnapshot.returnConfig = function(s3key, s3bucket, s3region, callback){
         Bucket: s3bucket
     };
     var buffer = "";
-    s3client.getObject(params, callback)
+    s3client.getObject(params)
         .createReadStream()
         .pipe(zlib.createGunzip())
         .on('data', function(chunk) {
--- a/node/cloudtrail_enabled_all_regions-periodic.js
+++ b/node/cloudtrail_enabled_all_regions-periodic.js
@@ -71,7 +71,7 @@ var readSnapshot = function(s3key, s3bucket, callback) {
 // where Config is set up to deliver
 readSnapshot.returnConfig = function(s3key, s3bucket, s3region, callback){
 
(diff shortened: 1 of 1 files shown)


node/cloudtrail_enabled_all_regions-periodic.js
```diff
@@ -3,61 +3,76 @@
...
+            readSnapshot.returnConfig(s3key, s3bucket, s3region, callback);
+        }
+    });
+};
+
 // Reads and parses the ConfigurationSnapshot from the S3 bucket
 // where Config is set up to deliver
- 
-function readSnapshot(s3client, s3key, s3bucket, callback) {
+readSnapshot.returnConfig = function(s3key, s3bucket, s3region, callback){
```
aaronkhoo: This convention ("readSnapshot.returnConfig") is a little unusual. So returnConfig is being scoped to readSnapshot?

aaronkhoo:
Hi @Sam-Martin, any thoughts on the question above?

Sam-Martin:
Hi @aaronkhoo, sorry, didn't get a chance to look at this yesterday! 
My original intention was to try and split out the readSnapshot function to be a bit more readable while still keeping it as the same object (as the bit I was splitting out wasn't terribly re-usable).
However, on further reflection it's a pretty overwrought solution and actually the same thing can be achieved with the latest commit I've just added.  
JavaScript isn't my strongest language so I'm more than happy to defer to your recommendation on the best way to implement this.

aaronkhoo:
The new code looks fine. Merging.

PR #32, opened 2016-06-17
Adding config rule to ensure presence and values of tags

travb:
This config rule will allow you to ensure not only the presence of defined tags (as the AWS Managed Rule), but also will let you specify a finite list of valid values for each tag.  This can be useful for enforcing that a valid CostCenter or Environment name is supplied, as arbitrary values can be used to be in compliance when only presence of tags is evaluated.


aaronkhoo:
Looks good. Merging.

PR #34, opened 2016-08-23
Update ec2_require_tags_with_valid_values.py

cornickj:
The change to the find_violation function adds checking for wildcards in values to allow Config to validate that a tag exists and has A value, not necessarily a specific value.  Useful for freeform attributes such as Owner, Creator, or Project that may have very dynamic value sets.


aaronkhoo:
Looks good. Merging.

PR #36, opened 2016-08-27
New config rule written in python

reithj:
New config rule flagging potential internet access as non-compliant. Has optional SNS support to trigger actions after evaluation.

### DIFF
--- a/python/ec2_no_internet_access-triggered.py
+++ b/python/ec2_no_internet_access-triggered.py
@@ -1,0 +1,185 @@
+#
+# This file made available under CC0 1.0 Universal (https://creativecommons.org/publicdomain/zero/1.0/legalcode)
+#
+# Ensures that there is no internet connectivity
+# Description: checks the given resource on potential internet access
+#
+# Trigger Type: Change Triggered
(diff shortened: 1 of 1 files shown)


python/ec2_no_internet_access-triggered.py
```diff
@@ -0,0 +1,185 @@
...
+  print('Compliance evaluation for %s: %s' % (configurationItem['resourceId'], compliance))
+ 
+  # inform config rules about our evaluation result
+  rule.put_evaluations(compliance, event['resultToken'])
+
+  event_detail = configurationItem
+  event_detail['evaluationResult'] = compliance
+
+  # publish a message with the evaluation result to SNS if a topic has been configured
+  if SNS_TOPIC != 'INSERT_SNS_TOPIC_ARN':
```
aaronkhoo: Most of the rule looks good. I wanted to point out that this is actually unnecessary. If you specify an SNS Topic in the AWS Config settings, Config Rules (both Custom and Managed) will send their evaluation results to that SNS topic each time PutEvaluations is called.

reithj:
Thanks for the hint, I removed SNS support

aaronkhoo:
Looks good. Merging......

PR #41, opened 2016-11-01
Ensure that no EC2 is publicly accessible except 80 and 443.

droidlabour:
These new rules are added

1. Check that no security group allows public access to ports other then 80 and 443
2. Check that no EC2 are in Public Subset
3. Check that no RDS are in Public Subset

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -275,3 +275,11 @@ Description: Checks that that the ingress permissions on an EC2 Security Group a
 
 Trigger Type: ```Change Triggered```
 Required Parameter: ```None```
+
+### 32. Ensure that no EC2 is publicly accessible except 80 and 443.
+Description: Check that no security groups allow public access to the ports other then 80 and 443.
+
+	python/ec2_sg_public_ingress_excluding_80_443.py
+
+Trigger Type: ```Change Triggered```
+Required Parameter: ```None```
--- a/RULES.md
+++ b/RULES.md
@@ -291,3 +291,11 @@ Description: Check that no EC2 are in Public Subnet.
 
 Trigger Type: ```Change Triggered```
 Required Parameter: ```None```
+
+### 34. Check that no RDS are in Public Subnet.
+Description: Check that no RDS are in Public Subnet.
+
+	python/rds_vpc_public_subnet.py
+
+Trigger Type: ```Change Triggered```
+Required Parameter: ```None```
--- a/RULES.md
+++ b/RULES.md
@@ -276,10 +276,18 @@ Description: Checks that that the ingress permissions on an EC2 Security Group a
 Trigger Type: ```Change Triggered```
 Required Parameter: ```None```
 
-### 32. Ensure that no EC2 is publicly accessible except 80 and 443.
+### 32. Check that no EC2 is publicly accessible except 80 and 443.
 Description: Check that no security groups allow public access to the ports other then 80 and 443.
 
 	python/ec2_sg_public_ingress_excluding_80_443.py
+
+Trigger Type: ```Change Triggered```
+Required Parameter: ```None```
+
+### 32. Check that no EC2 are in Public Subnet.
+Description: Check that no EC2 are in Public Subnet.
+
+	python/ec2_vpc_public_subnet.py
 
 Trigger Type: ```Change Triggered```
 Required Parameter: ```None```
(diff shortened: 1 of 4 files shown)


python/ec2_vpc_public_subnet.py
```diff
@@ -0,0 +1,95 @@
...
+log = logging.getLogger()
+log.setLevel(logging.INFO)
+
+def evaluate_compliance(configuration_item):
+    subnet_id   = configuration_item["configuration"]["subnetId"]
+    vpc_id      = configuration_item["configuration"]["vpcId"]
+    client      = boto3.client("ec2");
+    private     = True
+
+    response    = client.describe_route_tables()
```
ryflood: have you considered adding filters here to limit the scope of your request? You can add filters based on the subnet_id, vpc_id, and even the DestinationCidrBlock: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeRouteTables.html

RULES.md
```diff
@@ -275,3 +275,27 @@ Description: Checks that that the ingress permissions on an EC2 Security Group a
 
 Trigger Type: ```Change Triggered```
 Required Parameter: ```None```
+
+### 32. Check that no EC2 is publicly accessible except 80 and 443.
```
ryflood: Please consider updating the descriptions to be "EC2 Instances" where applicable, instead of just "EC2"

RULES.md
```diff
@@ -275,3 +275,27 @@ Description: Checks that that the ingress permissions on an EC2 Security Group a
...
+
+### 33. Check that no EC2 are in Public Subnet.
+Description: Check that no EC2 are in Public Subnet.
+
+	python/ec2_vpc_public_subnet.py
+
+Trigger Type: ```Change Triggered```
+Required Parameter: ```None```
+
+### 34. Check that no RDS are in Public Subnet.
```
ryflood: Please update the descriptions to be "RDS Instances" where applicable, instead of only "RDS"

RULES.md
ryflood: Please update the descriptions to be "EC2 Instances" where applicable, instead of only "EC2"

python/ec2_vpc_public_subnet.py
ryflood: Have you considered adding filters here to limit the scope of your response? You can add filters based on the subnet_id, vpc_id, and even the DestinationCidrBlock: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeRouteTables.html

RULES.md
droidlabour: Changes added in RULES.md https://github.com/PriceBoardIn/aws-config-rules/commit/8c06c87a0922742080fd3048ce37fbaf9a98ee60

RULES.md
droidlabour: Changes added in RULES.md https://github.com/PriceBoardIn/aws-config-rules/commit/8c06c87a0922742080fd3048ce37fbaf9a98ee60

droidlabour: Changes are also added in the relevant python files in comments sections. 
https://github.com/PriceBoardIn/aws-config-rules/commit/bed2c91449a6029b15e01e1db99f87be5c23159b
https://github.com/PriceBoardIn/aws-config-rules/commit/bbb28cf47c9d5cb8f2d2d3216a6c7ae7a1fe0626
https://github.com/PriceBoardIn/aws-config-rules/commit/f9de1356c4b743bbfa5e4e124a4400e700398c17

python/ec2_vpc_public_subnet.py
droidlabour: I've added describe_route_tables to filter with route.destination-cidr-block having 0.0.0.0/0
 https://github.com/PriceBoardIn/aws-config-rules/commit/b67f0a687adc0d4465144162a722e5eb238f4833

droidlabour:
@ryflood I've also added a condition to mark the subnet as public if there's Internet Gateway attached to the Route Tables route.
https://github.com/PriceBoardIn/aws-config-rules/commit/e46f858d44411105b0724a3068b83b38fa713499
https://github.com/PriceBoardIn/aws-config-rules/commit/78b6b78cd41217d9ca210a84857bc6a5ce1bf598

droidlabour:
@ryflood There was a bug related to distinguishing public vs private subnets.
I've changed it to only check based on the routes, i.e. only mark it as public if Internet Gateway exists otherwise mark it as private.
Also The filter for describe_route_tables is removed since filtering based on 0.0.0.0/0 won't return all the routes which will be the case if the route is private without.
https://github.com/PriceBoardIn/aws-config-rules/commit/07e9088d5cd921e105581f4d594eba33b26c0490

PR #53, opened 2017-07-28
Adding config rule to check exposed buckets

wtorrealba:
# WHAT

Adding config rule to check changes tracked by AWS Config related to S3 buckets that could potentially expose data

# WHY

We have seen that given some cases of data exposure people is looking for option to cover this kind of possible scenarios, AWS Config monitor for changes in the settings but do not have a default rule to look after changes in configuration on S3 resources.

# HOW

Based on the information that was shared the last few weeks this config rule is instead to check is a bucket is made public, basically it checks that the permissions do not include AuthenticatedUsers or AllUsers on the ACL level plus checking that the Bucket Policy does not have a Principal=* with effect=Allow.

At this time this does not include:
* Automatic remediation
* Validation on the object level (it is really expensive to check every object in a bucket)

# HOW TO TEST

Deploy the config rule,
Create a bucket (not exposed),
Validate the rule,
Expose the bucket by changing the ACL,
Validate the rule,
Fix the exposure,
Validate the rule,
Expose it again by changing the bucket policy
Validate the rule

# Considerations

### DIFF
--- a/python/s3-exposed-bucket.py
+++ b/python/s3-exposed-bucket.py
@@ -1,0 +1,110 @@
+#
+# This file made available under CC0 1.0 Universal
+# (https://creativecommons.org/publicdomain/zero/1.0/legalcode)
+#
+# Ensure that no S3 bucket allow public access.
+# Description: Checks that all S3 buckets block access on the ACL and bucket
+# policy level.
+#
+# Trigger Type: Change Triggered
+# Scope of Changes: S3:Bucket
+
+
+import json
+import boto3
+
+
+APPLICABLE_RESOURCES = ["AWS::S3::Bucket"]
+
+
+def validate_policy(policy_text):
+    if policy_text is None:
+        # The Bucket Policy is not set and therefore cannot be validated
+        return False
+    policy = json.loads(policy_text)
+    for s in policy["Statement"]:
+        if s["Effect"] == "Allow" and s["Principal"] == "*":
+            return True
+    return False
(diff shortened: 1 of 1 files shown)


nathaniela:
Getting the following error:

> 'ruleParameters': KeyError
Traceback (most recent call last):
File "/var/task/s3-exposed-bucket.py", line 90, in lambda_handler
rule_parameters = json.loads(event["ruleParameters"])
KeyError: 'ruleParameters'

I also printed the event and "ruleParameters" is not included in the event.

nathaniela:
Figured it out.
You have to provide at least one parameter key and one parameter value (input_parameters).
It doesn't really matter which parameters you send, the Lambda doesn't validate against it.

For example: I am sending **key:Validate, value:True** and now the function works fine.

BTW, 
The Lambda function iam policy looks, at a minimum, like this:

```
> {
>   "Version": "2012-10-17",
>   "Statement": [
>     {
>       "Effect": "Allow",
>       "Action": [
>         "logs:CreateLogGroup",
>         "logs:CreateLogStream",
>         "logs:PutLogEvents"
>       ],
>       "Resource": "arn:aws:logs:*:*:*"
>     },
>     {
>       "Effect": "Allow",
>       "Action": [
>         "s3:Get*",
>         "s3:List*",
>         "config:PutEvaluations"
>       ],
>       "Resource": "*"
>     }
>   ]
> }
```

PR #54, opened 2017-08-17
Pravinva patch 1

pravinva:
AWS Proserve Mission500  of config rules backlog

Verify VPC endpoint for S3 is configured in the route tables associated with the subnets


pravinva:
I have created another pull request because there were issues in squashing the old one.

PR #59, opened 2017-09-29
Added a fix for mfa rule to only evaluate console users

umerebryx:
This rule was evaluating cli users which don't have a console password and hence cannot use MFA login mechanism. Added a check to fetch user profile before deciding on the MFA attached status. For cli user's get_login_profile will throw an exception which we can use to skip over these users.

### DIFF
--- a/python/iam-mfa.py
+++ b/python/iam-mfa.py
@@ -10,7 +10,7 @@
 
 import json
 import boto3
-
+from botocore.exceptions import ClientError
 
 APPLICABLE_RESOURCES = ["AWS::IAM::User"]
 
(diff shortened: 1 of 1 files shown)


python/iam-mfa.py
```diff
@@ -20,8 +20,12 @@ def evaluate_compliance(configuration_item):
         return "NOT_APPLICABLE"
 
     user_name = configuration_item["configuration"]["userName"]
-
+
```
gregjinkim: Can you add this line at the top

from botocore.exceptions import ClientError

python/iam-mfa.py
```diff
@@ -20,8 +20,12 @@ def evaluate_compliance(configuration_item):
         return "NOT_APPLICABLE"
 
     user_name = configuration_item["configuration"]["userName"]
-
+    
     iam = boto3.client("iam")
+    try:
+        iam.get_login_profile(UserName=user_name)
+    except:
```
gregjinkim: except ClientError as e:
     if 'NoSuchEntity' in e.response['Error']['Code']:
        return "COMPLAINT"
     else:
        raise e

umerebryx:
@gregjinkim I have added the changes you asked.

netflash:
@gregjinkim any chances to merge this PR ?

PR #61, opened 2017-10-04
default security groups should not be attached

pravinva:
AWS Proserve Mission 500, config rules:

Default security groups should not be attached to EC2,RDS,ELB

### DIFF
--- a/python/default-security-group-attached
+++ b/python/default-security-group-attached
@@ -1,0 +1,81 @@
+# This file made available under CC0 1.0 Universal (https://creativecommons.org/publicdomain/zero/1.0/legalcode)
+# 
+# Description: Check that default security groups are not attached to EC2,RDS,ELB.
+#
+# Trigger Type: Change Triggered
+# Scope of Changes: EC2:NetworkInterfaces
+# Accepted Parameters: None
+# Your Lambda function execution role will need to have a policy that provides the appropriate
+# permissions.  Here is a policy that you can consider.  You should validate this for your own
+# environment
+#{
+#    "Version": "2012-10-17",
+#    "Statement": [
+#        {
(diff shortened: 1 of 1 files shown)


msambol:
Hi @pravinva. One of my customers is interested in this too, but also when the group name starts with `launch-wizard`. Is that something I can add? Or should I create another rule?

python/default-security-group-attached
```diff
@@ -0,0 +1,81 @@
+# This file made available under CC0 1.0 Universal (https://creativecommons.org/publicdomain/zero/1.0/legalcode)
+# 
+# Description: Check that default security groups are not attached to EC2,RDS,ELB.
+#
+# Trigger Type: Change Triggered
+# Scope of Changes: EC2:NetworkInterfaces
```
msambol: This should be singular --> `EC2:NetworkInterfaces`

python/default-security-group-attached
```diff
@@ -0,0 +1,81 @@
...
+#            "Resource": "*"
+#        }
+#    ]
+#}
+
+
+import boto3
+import json
+
+APPLICABLE_RESOURCES = ["AWS::EC2::NetworkInterfaces"]
```
msambol: More importantly, here.

pravinva:
@gregjinkim 
I wanted to check security groups attached to ELB as well, but now i realize its not in the 'applicable resources'. I'll use the groupname from security groups and resubmit. thanks

python/default-security-group-attached
pravinva: @msambol  thanks. will fix it

pravinva:
Closing pull request. Comments have been incorporated in https://github.com/awslabs/aws-config-rules/pull/73

PR #62, opened 2017-10-22
Check if any s3 bucket has bucket policy and if it does mark it non-compliant

droidlabour:


droidlabour:
@gregjinkim I've renamed it to `s3_bucket_policy_prohibited.py` Good catch.
Also I've changed to use Configuration Item instead of boto3. Apparently I wasn't aware of it.

droidlabour:
@gregjinkim Commit e7fae5d - 07e9088 were from my previous PR https://github.com/awslabs/aws-config-rules/pull/41/commits
I've added my latest commit on top of HEAD this repo.

python/s3_bucket_policy_prohibited.py
```diff
@@ -0,0 +1,75 @@
...
+import boto3
+import json
+import logging
+
+log = logging.getLogger()
+log.setLevel(logging.DEBUG)
+
+
+def evaluate_compliance(configuration_item):
+    bucket_policy = configuration_item["supplementaryConfiguration"].get("BucketPolicy")
```
gregjinkim: Can you handle cases of 
1) Configuration Items for resource types other than s3 buckets https://github.com/awslabs/aws-config-rules/blob/master/python/s3-exposed-bucket.py#L40
2) Configuration Items for deleted resources https://github.com/awslabs/aws-config-rules/blob/master/python/s3-exposed-bucket.py#L47
Both cases should be not_applicable

droidlabour: Added in 214072f

droidlabour:
@gregjinkim Thanks for merging. But It looks like I missed adding this new rule in the `RULES.md` doc. Can you please merge https://github.com/PriceBoardIn/aws-config-rules/commit/006708e59648a74c22cddcb11547ed7385d80cca

PR #63, opened 2017-10-31
EC2 instances without the latest SSM agent

ppratee:


python/ec2-ssm-version-check.py
```diff
@@ -0,0 +1,83 @@
...
+    else:
+        return {
+                'compliance_type': "NON_COMPLIANT",
+                'annotation': "SSM agent is not installed"
+               }
+                   
+def evaluation_result(instanceid):
+    ec2client = boto3.client('ec2')
+    instancedetail = ec2client.describe_instance_status(InstanceIds=[str(instanceid)])
+    if instancedetail['InstanceStatuses']:
```
gregjinkim: You can check if it is running by looking at the 'state' field in the Configuration Item

ppratee: This is a periodically triggered rule and I am not passing any configuration item to the lambda function. I generate a list of all ec2 instances running in the region using list-discovered-items api call within the lambda function and then iterate through each to get ssm version details. 
Is there another way to do this?

gregjinkim: If you made it a change triggered rule, the rule would receive a Configuration Item whenever an ec2 instance got created or changed, and you would be able to evaluate this rule for that instance at much closer to real-time than if you just have it be periodic. All the information that you need about the ec2 instance is in the configuration item I believe

ppratee: This may not solve the purpose of this rule. The purpose here is to track the version of the agent on instance periodically as AWS releases new SSM agent versions frequently. Hence a need to frequently compare the version running on instances with the latest version released by AWS. If I change the rule to change triggered type, the rule will get triggered only when there is a change in the instance which could be way less frequent than version release by AWS. In case there are no changes to an instance, the customer would not get to know which instances are running an outdated version despite multiple release cycles by AWS.

PR #64, opened 2017-10-31
Add ec2_launch_wizard_security_group_prohibited.py

msambol:
Add a custom Config Rule to check that security groups 
prefixed with "launch-wizard" are not associated with 
network interfaces.


msambol:
@gregjinkim Done.

PR #66, opened 2017-11-01
Fix false positive in *_vpc_public_subnet.py

msambol:
Need a second eye here but I believe the current
code produces a false positive. For example, what
if the main route table has a public route but the subnet 
is associated with a private route table?

### DIFF
--- a/python/ec2_vpc_public_subnet.py
+++ b/python/ec2_vpc_public_subnet.py
@@ -45,7 +45,6 @@ def evaluate_compliance(configuration_item):
     subnet_id   = configuration_item["configuration"]["subnetId"]
     vpc_id      = configuration_item["configuration"]["vpcId"]
     client      = boto3.client("ec2");
-    private     = True
 
     response    = client.describe_route_tables()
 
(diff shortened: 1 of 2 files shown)


msambol:
@droidlabour Wouldn't mind your review here since you authored the original. Thanks!

msambol:
@aaronkhoo Can you review and merge? One of my colleagues confirmed this bug too.

BetaKevin:
I don't think this will capture a certain use case.  Specifically with RDS instances as it needs multiple subnets to fail.

If the RDS instance has more than 1 subnetidentifier and the following is true:
- 1 or more subnets are private and not using the main table
- 1 subnet is public using the main table.  

In this case the code you have submitted will pass even though it should fail.

msambol:
@BetaKevin Fixed. Thanks!

msambol:
@aaronkhoo can you review? This fixes a bug with 2 rules.

jwasham:
Thanks for the fix, @msambol !

PR #67, opened 2017-11-02
Exposed security groups fails if Port Range = "All"

msambol:
If the source for a security group is 
`0.0.0.0/0` and the port range is `All`,
this rule will not detect the exposed group.

Also, it's good practice to check if the 
resource was deleted.


aaronkhoo:
Looks good. Merging request.

PR #69, opened 2017-11-03
Add checks to prohibit port range and protocol values of "All"

msambol:
This checks that inbound rules for security groups
are explicit in terms of protocol and port grange. 
A value of "All" will make a security group non-compliant.


msambol:
@aaronkhoo Can you review and merge this PR?

PR #72, opened 2017-11-14
Handled the case where an IAM User was deleted since the last evaluation.

aidanmelen:
Handled the case where an IAM User was deleted since the last evaluation. The next time the config rule is run, the configuration_item["configuration"] will be 'None' causing a runtime error.

### DIFF
--- a/python/iam-mfa.py
+++ b/python/iam-mfa.py
@@ -18,6 +18,9 @@ APPLICABLE_RESOURCES = ["AWS::IAM::User"]
 def evaluate_compliance(configuration_item):
     if configuration_item["resourceType"] not in APPLICABLE_RESOURCES:
         return "NOT_APPLICABLE"
+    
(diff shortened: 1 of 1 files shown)


python/iam-mfa.py
```diff
@@ -18,6 +18,9 @@
 def evaluate_compliance(configuration_item):
     if configuration_item["resourceType"] not in APPLICABLE_RESOURCES:
         return "NOT_APPLICABLE"
+    
+    if not configuration_item["configuration"]:
```
aaronkhoo: An explicit check on the configuration item status might be better here. See https://github.com/awslabs/aws-config-rules/blob/86c3f70bc56a7b2711319f0a6dc0b12855f5cb6f/python/ec2-exposed-group.py#L62 for an example of how to do this.

jongogogo:
No response. Closing

PR #73, opened 2017-11-14
Added rule for default-security-group-attached.py

pravinva:
Default security groups should not be attached to EC2


msambol:
@pravinva Can you squash these commits?

python/subnet-s3-vpc-endpoint.py
```diff
@@ -0,0 +1,102 @@
...
+import logging
+
+
+APPLICABLE_RESOURCES = ["AWS::EC2::SubnetId"]
+
+log = logging.getLogger()
+log.setLevel(logging.INFO)
+
+def evaluate_compliance(configuration_item):
+        if configuration_item["resourceType"] not in APPLICABLE_RESOURCES:
```
msambol: indentation seems off here.

python/subnet-s3-vpc-endpoint.py
```diff
@@ -0,0 +1,102 @@
...
+            "annotation" : 'SubnetId '+ str(subnet_id)+ '  does not have a S3 VPC endpoint'
+        }
+
+def lambda_handler(event, context):
+    log.debug('Event %s', event)
+    invoking_event      = json.loads(event['invokingEvent'])
+    configuration_item  = invoking_event["configurationItem"]
+    
+
+
```
msambol: can you clean up these newlines?

PR #77, opened 2017-12-18
Periodic Config Rules

ppratee:
Periodic Config Rules for:
1) Daily check the SSM agent version on all EC2 instances running in the region.
2) Daily check KMS CMK keys without automatic rotation enabled


python/ssm-compliance-check-lambda-fn.py
```diff
@@ -0,0 +1,110 @@
...
+    print(event)
+    invoking_event = json.loads(event['invokingEvent'])
+    configclient = boto3.client('config')
+    respond = configclient.list_discovered_resources(resourceType='AWS::EC2::Instance')
+    print(respond)
+    for resource in respond['resourceIdentifiers']:
+            instanceid = resource['resourceId']
+            evaluationresult = evaluation_result(instanceid) 
+            print(instanceid, evaluationresult)
+            response = configclient.put_evaluations(
```
gregjinkim: You could collect all the evaluations and then make one PutEvaluations API call (or batch them up into 100's)

ppratee:
I get what you are saying. I though of that earlier but could not find a way to send evaluation results for multiple instances in a single API call. Could you point me to such an example code?

netflash:
Hi there,
is this going anywhere ?

elsawyer:
@ppratee - Regarding putting multiple evaluation results in a single API call, PutEvaluations accepts a list of evaluations - multiple instances work fine with this since you specify a resourceId for each evaluation. For an example, you can check out [https://github.com/awslabs/aws-config-rules/blob/04e3ecef1b6846e969d88fabcfc34a05256d1326/python/ec2_require_ebs_snapshots_for_volumes#L71](url)

jongogogo:
No response. Closing

PR #78, opened 2017-12-19
ec2_sg_public_ingress_remove.py

skstussy:
Checks that all security groups block access to the specified ports and removes ip permissions that violate the rule.

### DIFF
--- a/python/ec2_sg_public_ingress_remove.py
+++ b/python/ec2_sg_public_ingress_remove.py
@@ -1,0 +1,165 @@
+#
+# This file made available under CC0 1.0 Universal (https://creativecommons.org/publicdomain/zero/1.0/legalcode)
+#
+# Ensure that no security groups allow public access to the specified ports.
+# Description: Checks that all security groups block access to the specified ports and removes ip permissions that violate the rule.
+#
+# Trigger Type: Change Triggered
+# Scope of Changes: EC2:SecurityGroup
+# Accepted Parameters: examplePort1, exampleRange1, examplePort2, ...
+# Example Value: 8080, 1-1024, 2375, ...
+
+
+import json
+import boto3
+import botocore
+
+
+
+APPLICABLE_RESOURCES = ["AWS::EC2::SecurityGroup"]
+
+def capitalize(x):
+   if isinstance(x, list):
+     return [capitalize(v) for v in x]
+   elif isinstance(x, dict):
+     return {k[0].upper() + k[1:]: capitalize(v) for k, v in x.items()}
+   else:
+     return x
+
+def expand_range(ports):
+    if "-" in ports:
+        return range(int(ports.split("-")[0]), int(ports.split("-")[1])+1)
+    else:
+        return [int(ports)]
+
+def revoke_permissions(permissions):
+    #global group_id
+    group_id = configuration_item["configuration"]["groupId"]
+    client = boto3.client("ec2");
+    print("revoking permissions in ", group_id)
+    #Capatilize the first letter of each key
+    revoke_permissions = [capitalize(permission)]
+    #Transform Ipv4Ranges -> IpRanges
+    for element in revoke_permissions: 
+        #dictionary[new_key] = dictionary.pop(old_key)
+        element["IpRanges"] = element.pop("Ipv4Ranges")
+        print("input for revoke permissions: " + str(revoke_permissions))
+        try:
+            client.revoke_security_group_ingress(GroupId=group_id, IpPermissions=revoke_permissions)
+            return {
+                "compliance_type" : "NON_COMPLIANT",
+                "annotation" : " revoking: " + str(revoke_permissions)
+            }
+        except botocore.exceptions.ClientError as e:
+            print (e)
+            return {
+                "compliance_type" : "NON_COMPLIANT",
+                "annotation" : "revoke_security_group_ingress error " + e
+            }
+                
+def find_exposed_ports(ip_permissions):
+    exposed_ports = []
+    global permission
+    for permission in ip_permissions or []:
+        for cidrIpv6 in permission["ipv6Ranges"]:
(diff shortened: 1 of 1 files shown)


python/ec2_sg_public_ingress_remove.py
```diff
@@ -0,0 +1,165 @@
...
+    revoke_permissions = [capitalize(permission)]
+    #Transform Ipv4Ranges -> IpRanges
+    for element in revoke_permissions: 
+        #dictionary[new_key] = dictionary.pop(old_key)
+        element["IpRanges"] = element.pop("Ipv4Ranges")
+        print("input for revoke permissions: " + str(revoke_permissions))
+        try:
+            client.revoke_security_group_ingress(GroupId=group_id, IpPermissions=revoke_permissions)
+            return {
+                "compliance_type" : "NON_COMPLIANT",
```
elsawyer: I don't think this case should be considered non compliant - permissions were revoked, so as I understand it, all is good, right?

Should still have the annotation to say what was revoked though.

python/ec2_sg_public_ingress_remove.py
```diff
@@ -0,0 +1,165 @@
...
+# This file made available under CC0 1.0 Universal (https://creativecommons.org/publicdomain/zero/1.0/legalcode)
+#
+# Ensure that no security groups allow public access to the specified ports.
+# Description: Checks that all security groups block access to the specified ports and removes ip permissions that violate the rule.
+#
+# Trigger Type: Change Triggered
+# Scope of Changes: EC2:SecurityGroup
+# Accepted Parameters: examplePort1, exampleRange1, examplePort2, ...
+# Example Value: 8080, 1-1024, 2375, ...
+
```
elsawyer: I'm guessing the lambda execution role needs specific permissions to execute this, what with the revoking of permissions.

Can you call that out here, and add a sample policy?

python/ec2_sg_public_ingress_remove.py
```diff
@@ -0,0 +1,165 @@
...
+                else:
+                    exposed_ports.extend(range(0,65535+1))
+                    
+    return exposed_ports
+
+
+def find_violation(ip_permissions, forbidden_ports):
+    exposed_ports = find_exposed_ports(ip_permissions)
+    for forbidden in forbidden_ports:
+        ports = expand_range(forbidden_ports[forbidden])
```
elsawyer: why are you indexing into forbidden_ports with one of the items in the list?

python/ec2_sg_public_ingress_remove.py
```diff
@@ -0,0 +1,165 @@
...
+def revoke_permissions(permissions):
+    #global group_id
+    group_id = configuration_item["configuration"]["groupId"]
+    client = boto3.client("ec2");
+    print("revoking permissions in ", group_id)
+    #Capatilize the first letter of each key
+    revoke_permissions = [capitalize(permission)]
+    #Transform Ipv4Ranges -> IpRanges
+    for element in revoke_permissions: 
+        #dictionary[new_key] = dictionary.pop(old_key)
```
elsawyer: I don't think this comment is needed

python/ec2_sg_public_ingress_remove.py
```diff
@@ -0,0 +1,165 @@
...
+        except botocore.exceptions.ClientError as e:
+            print (e)
+            return {
+                "compliance_type" : "NON_COMPLIANT",
+                "annotation" : "revoke_security_group_ingress error " + e
+            }
+                
+def find_exposed_ports(ip_permissions):
+    exposed_ports = []
+    global permission
```
elsawyer: This shouldn't need to be global

python/ec2_sg_public_ingress_remove.py
```diff
@@ -0,0 +1,165 @@
...
+    for permission in ip_permissions or []:
+        for cidrIpv6 in permission["ipv6Ranges"]:
+            if "::/0" in cidrIpv6["cidrIpv6"]:
```
elsawyer: Could you change the naming here? Maybe change to 
`for ipv6Range in permission["ipv6Ranges"]:`?
Looks kinda weird to have: `cidrIpv6["cidrIpv6"]`

python/ec2_sg_public_ingress_remove.py
skstussy: This is personal choice but it makes sense to make it COMPLIANT. You will probably never see the compliance state report non-compliant if you use COMPLIANT. I personally decided to use NON-COMPLIANT because I want to see when a security groups has been found to be non-compliant, even though the rule is revoking the offending ip permissions

python/ec2_sg_public_ingress_remove.py
skstussy: Permissions Required:

AWSConfigRulesExecutionRole (Required for Config service)
AWSLambdaBasicExecutionRole (Required for Lambda service)
Custom EC2 policy			(Allow function to describe SG and revoke IP Permssions Ingress)

{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "ec2:DescribeSecurityGroups",
                "ec2:RevokeSecurityGroupIngress"
            ],
            "Resource": "*"
        }
    ]
}

python/ec2_sg_public_ingress_remove.py
skstussy: yes 
 "#dictionary[new_key] = dictionary.pop(old_key)"
 can be removed

jongogogo:
Design following best practices: 
- split the evaluation and the reaction.
- Use CloudWatch Event + Lambda for the reaction
- Rule on IPv4 already exists: https://docs.aws.amazon.com/config/latest/developerguide/restricted-common-ports.html
- Rule on IPv6: Please do a new PR for this rule only.

Thanks

PR #80, opened 2017-12-28
Add rule to check for S3 bucket default encryption

msambol:
This rule checks S3 buckets for default encryption.

More on default encryption [here](http://docs.aws.amazon.com/AmazonS3/latest/dev/bucket-encryption.html).

### DIFF
--- a/python/s3_bucket_default_encryption_enabled.py
+++ b/python/s3_bucket_default_encryption_enabled.py
@@ -1,0 +1,106 @@
+# This file made available under CC0 1.0 Universal (https://creativecommons.org/publicdomain/zero/1.0/legalcode)
+#
+# Description: Check that S3 buckets have default encryption enabled.
+#
+# Trigger Type: Change Triggered
+# Scope of Changes: S3:Bucket
+# Accepted Parameters: None
+# Your Lambda function execution role will need to have a policy that provides
+# the appropriate permissions. Here is a policy that you can consider.
+# You should validate this for your own environment.
+#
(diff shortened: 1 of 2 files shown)


msambol:
@aaronkhoo Can you review please?

python/s3_bucket_default_encryption_enabled.py
```diff
@@ -0,0 +1,106 @@
...
+#            "Effect": "Allow",
+#            "Action": [
+#                "config:PutEvaluations"
+#            ],
+#            "Resource": "*"
+#        },
+#        {
+#            "Effect": "Allow",
+#            "Action": [
+#                "s3:Get*"
```
kuentzm: Action is to permissive.  Can it be scoped down to "s3:GetEncryptionConfiguration"

python/s3_bucket_default_encryption_enabled.py
```diff
@@ -0,0 +1,106 @@
...
+#            ],
+#            "Resource": "*"
```
kuentzm: Can resource be scoped down to "arn:aws:s3:::*"

msambol:
@kuentzm Fixed. Thanks!

msambol:
@kuentzm Think this is ready. Can you take another look?

kuentzm:
Yes, thanks!  @aaronkhoo will need to review

msambol:
@aaronkhoo this is ready for your review.

PR #82, opened 2018-02-13
Add Python Testing

kddejong:
- Include Tox to make running tests easier
- Fix a set of rules based on failures in the code
- Write tests for a set of linting rules


kddejong:
Lots of linting fixes for python still required.  When I have some more time I will fix that as well.

jongogogo:
I suggest to open a conversation first on how to integrate testing in this repo. It is a step in a good direction. 
For fixes in the rules, please re-submit them independently.

PR #94, opened 2018-08-08
New custom rule for checking patch compliance

lahirian:
Hi,

This is my first time submitting a pull request on this repo. Please let me know if I need to modify something. 

This is a new custom rule that I wrote to find out if a set of ec2 instances (filtered based on a given tag and value) is compliant to a SSM patch baseline.

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -331,3 +331,11 @@ Description: Check that S3 buckets have default encryption enabled.
 
 Trigger Type: ```Change Triggered```
 Required Parameter: ```None```
+
+### 38. Check the patch compliance of a set of instances based on their tag and SSM 
+Description: Check the patch compliance of a set of instances based on their tags
+
+	python/ssm_patch_compliance.py
+
+Trigger Type: ```Periodic```
+Required Parameter: ```Tag Key: Tag Value```
--- a/python/ssm_patch_compliance.py
+++ b/python/ssm_patch_compliance.py
@@ -1,0 +1,142 @@
+#
+# This file made available under CC0 1.0 Universal (https://creativecommons.org/publicdomain/zero/1.0/legalcode)
+#
+# Run SSM patch compliance check against ec2 instances
+# Description: Checks ec2 instances are patch compliant via SSM against SSM patch baseline.
+#
+# Trigger Type: Periodic
+# Scope of Changes: N/A
+# Required Parameter name: PoliciesToCheck
+# Required Parameter value example: Name bastion (Tag key and value for EC2 instances). It will find all instances with that particular
+# tag and for each instance it will insert a compliant/non compliant evaluation into AWS config.
+# IAM role running this lambda should be able to call ec2:describe , ssm:list_compliance_items and config:putEvaluations
+
+import json
+import boto3
+
+
+
+
+APPLICABLE_RESOURCES = ["AWS::EC2::Instance"]
+
+
+def validate_patch_baseline(instanceid):
+    client = boto3.client('ssm')
+    filters = [
+		{'Key':'ComplianceType','Values':['Patch'],'Type':'EQUAL'},
+		{'Key':'Status','Values':['NON_COMPLIANT'],'Type':'EQUAL'}
+	     ]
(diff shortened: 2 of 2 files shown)


python/ssm_patch_compliance.py
```diff
@@ -0,0 +1,142 @@
...
+        "annotation": "This resource is compliant with the rule."
+    }
+
+
+
+def lambda_handler(event, context):
+
+    print(event)
+    invoking_event = json.loads(event["invokingEvent"])
+    #configuration_item = invoking_event["configurationItem"]
```
jongogogo: Why? This rule should be configuration changes triggered, as far I understand the intend.

python/ssm_patch_compliance.py
```diff
@@ -0,0 +1,142 @@
...
+    configuration_item = {'resourceType':'AWS::EC2::Instance'}
```
jongogogo: Should not be hardcoded

python/ssm_patch_compliance.py
```diff
@@ -0,0 +1,142 @@
...
+import json
+import boto3
+
+
+
+
+APPLICABLE_RESOURCES = ["AWS::EC2::Instance"]
+
+
+def validate_patch_baseline(instanceid):
```
jongogogo: Split this function in two which return a boolean: 1) checking the presence of an agent, 2) checking the presence of an unpatched instance.

python/ssm_patch_compliance.py
```diff
@@ -0,0 +1,142 @@
...
+    client = boto3.client('ssm')
+    filters = [
+		{'Key':'ComplianceType','Values':['Patch'],'Type':'EQUAL'},
+		{'Key':'Status','Values':['NON_COMPLIANT'],'Type':'EQUAL'}
+	     ]
+    response = client.list_compliance_items(Filters=filters,ResourceIds=[instanceid],ResourceTypes=['ManagedInstance'])
+    print(response)
+    if len(response['ComplianceItems']) > 0:
+        #link_to_ssm = '<a href="https://console.aws.amazon.com/systems-manager/managed-instances/' + instanceid + '/configurationCompliance>"'
```
jongogogo: Remove

python/ssm_patch_compliance.py
```diff
@@ -0,0 +1,142 @@
...
+        security_critical = ''
+        
+        for item in response['ComplianceItems']:
+            
+            if 'Classification' in item['Details']:
+                if item['Details']['Classification'] == 'Security':
+                    #security_critical = security_critical + item['Title'] + ":"
```
jongogogo: Remove

python/ssm_patch_compliance.py
```diff
@@ -0,0 +1,142 @@
...
+                    security_critical = " Several pending security patches."
+        violation="NON_COMPLIANT: " + instanceid  + ":" + security_critical
+        print(violation)
+        return violation
+
+    response = client.describe_instance_patch_states(InstanceIds=[instanceid])
+    
+    if len(response['InstancePatchStates']) == 0:
```
jongogogo:
```
if not response['InstancePatchStates']:
```

RULES.md
```diff
@@ -331,3 +331,11 @@ Description: Check that S3 buckets have default encryption enabled.
 
 Trigger Type: ```Change Triggered```
 Required Parameter: ```None```
+
+### 38. Check the patch compliance of a set of instances based on their tag and SSM
```
jongogogo: Please remove this file from your PR. We are working on getting this file auto-generated.

jongogogo:
As been launched as managed rules: https://docs.aws.amazon.com/config/latest/developerguide/ec2-managedinstance-patch-compliance-status-check.html

lahirian:
Thanks for the review. Sorry I just noticed this and missed replying on previous comments.

python/ssm_patch_compliance.py
lahirian: At the same time there may be occasions when we want to run the config rule manually to ensure that all instances are patch compliant ie in event of some major security vulnerability announcement. But I think this gives a good foundation for users to customize as needed.

jongogogo:
No worries. Keep pushing PR :)

PR #112, opened 2018-09-19
Add IAM policy required config rule

HieronymusLex:

*Issue #, if available:*

*Description of changes:*
Adds a Config Rule that verifies all IAM Roles/Users have the given policy attached to them (either directly or through groups). Includes ability to exclude certain roles/users from the rule

### DIFF
--- a/python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
+++ b/python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
@@ -30,14 +30,12 @@ ASSUME_ROLE_MODE = False
 
 
 def should_ignore_config_item(config_item, ignored_roles, ignored_users):
-    return (config_item['configurationItemStatus'] == 'ResourceDeleted')\
-            or (config_item['resourceType'] == 'AWS::IAM::Role' and config_item['resourceName'] in ignored_roles)\
-            or (config_item['resourceType'] == 'AWS::IAM::User' and config_item['resourceName'] in ignored_users)\
+    return (config_item['resourceType'] == 'AWS::IAM::Role' and config_item['resourceName'] in ignored_roles) \
+            or (config_item['resourceType'] == 'AWS::IAM::User' and config_item['resourceName'] in ignored_users) \
             or (config_item['ARN'].rsplit("/")[1] == 'aws-service-role')
 
 
-def get_user_policies(username):
-    client = boto3.client('iam')
+def get_user_policies(username, client):
     # Get the users managed policies
     policies = [
         policy['PolicyName'] for policy in client.list_attached_user_policies(UserName=username)['AttachedPolicies']
@@ -54,19 +52,18 @@ def get_user_policies(username):
     return policies
 
 
-def get_role_policies(role_name):
-    client = boto3.client('iam')
+def get_role_policies(role_name, client):
     # Get the users managed policies
     return [
         policy['PolicyName'] for policy in client.list_attached_role_policies(RoleName=role_name)['AttachedPolicies']
     ]
 
 
-def has_policy_attached(resource_name, resource_type, policy_name):
+def has_policy_attached(resource_name, resource_type, policy_name, client):
     if resource_type == 'AWS::IAM::User':
-        return policy_name in get_user_policies(resource_name)
+        return policy_name in get_user_policies(resource_name, client)
     elif resource_type == 'AWS::IAM::Role':
-        return policy_name in get_role_policies(resource_name)
+        return policy_name in get_role_policies(resource_name, client)
     else:
         raise ValueError('Unable to handle resource type {}'.format(resource_type))
 
@@ -93,10 +90,11 @@ def evaluate_compliance(event, configuration_item, valid_rule_parameters):
     ignored_roles = valid_rule_parameters["ignored_roles"]
     ignored_users = valid_rule_parameters["ignored_users"]
     policy_name = valid_rule_parameters['policy_name']
+    client = get_client('iam', event)
 
     if should_ignore_config_item(configuration_item, ignored_roles, ignored_users):
         return 'NOT_APPLICABLE'
-    elif has_policy_attached(configuration_item['resourceName'], configuration_item['resourceType'], policy_name):
+    elif has_policy_attached(configuration_item['resourceName'], configuration_item['resourceType'], policy_name, client):
         return 'COMPLIANT'
     else:
         return 'NON_COMPLIANT'
--- a/python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
+++ b/python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
@@ -8,8 +8,95 @@
 # or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for
 # the specific language governing permissions and limitations under the License.
+"""
+#####################################
+##           Gherkin               ##
+#####################################
+ Rule Name:
+   IAM_POLICY_REQUIRED
+ Description:
+   To check IAM users and roles have a given policy attached directly or through a group.
+ Trigger:
+   Configuration Change on AWS::IAM::User/AWS::IAM::Role
+ Reports on:
+   AWS::IAM::User,AWS::IAM::Role
+ Rule Parameters:
+   | ---------------------- | --------- | -------------------------------------------------------- |
+   | Parameter Name         | Type      | Description                                              |
+   | ---------------------- | --------- | -------------------------------------------------------- |
+   | policyArns             | Required  | ARNs of the policies which should be attached to all     |
+   |                        |           | users and roles.                                         |
+   | ---------------------- | --------- | -------------------------------------------------------- |
+   | exceptionList          | Optional  | Represents the IAM users and roles which are exempted    |
+   |                        |           | from the IAM Config rule. The valid users in this list   |
+   |                        |           | will be compliant by default.                            |
+   |                        |           | List of all the User Names separated by a comma          |
+   | ---------------------- | --------- | -------------------------------------------------------- |
+ Feature:
+   As: a Security Officer
+   I want: To ensure that IAM roles and users have mandatory policies attached
+   In order to: enforce mandatory permissions for IAM entities
+ Scenarios:
+  Scenario: 1
+     Given: No IAM Users or Roles exist
+      When: Evaluation occurs
+      Then: Return "NOT_APPLICABLE"
+   Scenario: 2
+     Given: exceptionList is configured
+       And: A user listed in the exceptionList is not an alphanumerical string
+      When: Evaluation occurs
+      Then: Return an error
+   Scenario: 3
+     Given: exceptionList is configured
+       And: A role listed in the exceptionList is not an alphanumerical string
+      When: Evaluation occurs
+      Then: Return an error
+   Scenario: 4
+     Given: policyArns is configured
+       And: policyArns does not contain valid ARNs
+      Then: Return an error
+   Scenario: 5
+     Given: An IAM user exists
+       And: exceptionList is configured and valid
+       And: The IAM user is listed in the exceptionList
+      When: Evaluation occurs
+      Then: Return COMPLIANT
+   Scenario: 6
+     Given: An IAM role exists
+       And: exceptionList is configured and valid
+       And: The IAM role is listed in the exceptionList
+      When: Evaluation occurs
+      Then: Return COMPLIANT
+   Scenario: 7
+     Given: An IAM user exists
+       And: The user has all the policies listed in policyArns attached
+      When: Evaluation occurs
+      Then: Return COMPLIANT
+   Scenario: 8
+     Given: An IAM user exists
+       And: The IAM user groups combined have the policies listed in policyArns attached
+      When: Evaluation occurs
+      Then: Return COMPLIANT
+   Scenario: 9
+     Given: An IAM role exists
+       And: The role has all the policies listed in policyArns attached
+      When: Evaluation occurs
+      Then: Return COMPLIANT
+   Scenario: 10
+     Given: An IAM user
+       And: The IAM user does not have all the polciies listed in policyArns attached
+       And: The IAM user's groups combined do not have all the policy listed in policyArns attached
+      When: Evaluation occurs
+      Then: return NON_COMPLIANT
+   Scenario: 11
+     Given: An IAM role
+       And: The IAM role does not have all the policies listed in policyArns attached
+      When: Evaluation occurs
+      Then: return NON_COMPLIANT
+"""
 
 import json
+import re
 import datetime
 import boto3
 import botocore
@@ -35,35 +122,45 @@ def should_ignore_config_item(config_item, ignored_roles, ignored_users):
             or (config_item['ARN'].rsplit("/")[1] == 'aws-service-role')
 
 
-def get_user_policies(username, client):
+def get_attached_policies(configuration_item):
     # Get the users managed policies
-    policies = [
-        policy['PolicyName'] for policy in client.list_attached_user_policies(UserName=username)['AttachedPolicies']
-    ]
-    # Get the policies from the users groups
-    groups = [
-        group['GroupName'] for group in client.list_groups_for_user(UserName=username)['Groups']
-    ]
-    for group in groups:
-        policies.extend([
-            policy['PolicyName'] for policy in client.list_attached_group_policies(GroupName=group)['AttachedPolicies']
-        ])
-
-    return policies
-
-
-def get_role_policies(role_name, client):
-    # Get the users managed policies
-    return [
-        policy['PolicyName'] for policy in client.list_attached_role_policies(RoleName=role_name)['AttachedPolicies']
+    attach_policies = [
+        policy["policyArn"] for policy in configuration_item["configuration"].get("attachedManagedPolicies")
     ]
 
+    return attach_policies
 
-def has_policy_attached(resource_name, resource_type, policy_name, client):
+
+def list_contains_all(source_list, items):
+    return all(policy in source_list for policy in items)
+
+
+def paginate(client, method, **kwargs):
+    paginator = client.get_paginator(method.__name__)
+    for page in paginator.paginate(**kwargs).result_key_iters():
+        for result in page:
+            yield result
+
+
+def has_policy_attached(event, configuration_item, policy_arns):
+    resource_type = configuration_item['resourceType']
     if resource_type == 'AWS::IAM::User':
-        return policy_name in get_user_policies(resource_name, client)
+        managed_policies = get_attached_policies(configuration_item)
+        if list_contains_all(managed_policies, policy_arns):
+            return True
+        # Additively check the users groups to see if they have the required policies
+        client = get_client('iam', event)
+        groups = configuration_item["configuration"].get("groupList", [])
+        for group in groups:
+            attached_policies = paginate(client, client.list_attached_group_policies, **{'GroupName': group})
+            for policy in attached_policies:
+                managed_policies.append(policy['PolicyArn'])
+                if list_contains_all(managed_policies, policy_arns):
+                    return True
+
+        return False
     elif resource_type == 'AWS::IAM::Role':
-        return policy_name in get_role_policies(resource_name, client)
+        return list_contains_all(get_attached_policies(configuration_item), policy_arns)
     else:
         raise ValueError('Unable to handle resource type {}'.format(resource_type))
 
@@ -87,17 +184,31 @@ def evaluate_compliance(event, configuration_item, valid_rule_parameters):
     2 -- if a None or a list of dictionary is returned, the old evaluation(s) which are not returned in the new evaluation list are returned as NOT_APPLICABLE by the Boilerplate code
     3 -- if None or an empty string, list or dict is returned, the Boilerplate code will put a "shadow" evaluation to feedback that the evaluation took place properly
     """
-    ignored_roles = valid_rule_parameters["ignoredRoles"]
-    ignored_users = valid_rule_parameters["ignoredUsers"]
-    policy_name = valid_rule_parameters['policyName']
-    client = get_client('iam', event)
+    policy_arns = valid_rule_parameters['policyArns']
+    exception_list = valid_rule_parameters["exceptionList"]
+    ignored_roles = exception_list["roles"]
+    ignored_users = exception_list["users"]
 
     if should_ignore_config_item(configuration_item, ignored_roles, ignored_users):
-        return 'COMPLIANT'
-    elif has_policy_attached(configuration_item['resourceName'], configuration_item['resourceType'], policy_name, client):
-        return 'COMPLIANT'
+        return build_evaluation_from_config_item(configuration_item, 'COMPLIANT', 'Ignored IAM entity')
+    elif has_policy_attached(event, configuration_item, policy_arns):
+        return build_evaluation_from_config_item(configuration_item, 'COMPLIANT', 'All expected policies attached')
     else:
-        return 'NON_COMPLIANT'
+        return build_evaluation_from_config_item(configuration_item, 'NON_COMPLIANT', 'IAM entity missing policies')
+
+
+def is_valid_arn(arn):
+    pattern = re.compile("arn:(aws[a-zA-Z-]*)?:iam::(aws|\d{12}):policy\/[a-zA-Z0-9-_\/]+")
+    return pattern.match(arn)
+
+
+def extract_entities_from_exception_list(entity_type, exception_list):
+    pattern = re.compile("{Type}:\s?\[([a-zA-Z0-9-_,]+)\]".format(Type=entity_type))
+    matches = pattern.search(exception_list)
+    if matches:
+        return matches.group(1).split(",")
+    else:
+        return []
 
 
 def evaluate_parameters(rule_parameters):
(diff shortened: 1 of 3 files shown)


python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,405 @@
...
+
+def should_ignore_config_item(config_item, ignored_roles, ignored_users):
+    return (config_item['configurationItemStatus'] == 'ResourceDeleted')\
+            or (config_item['resourceType'] == 'AWS::IAM::Role' and config_item['resourceName'] in ignored_roles)\
+            or (config_item['resourceType'] == 'AWS::IAM::User' and config_item['resourceName'] in ignored_users)\
+            or (config_item['ARN'].rsplit("/")[1] == 'aws-service-role')
+
+
+def get_user_policies(username):
+    client = boto3.client('iam')
```
jongogogo: Use get_client() function to enable this rule to be used in a cross-account architecture

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,405 @@
...
+    # Get the users managed policies
+    policies = [
+        policy['PolicyName'] for policy in client.list_attached_user_policies(UserName=username)['AttachedPolicies']
```
jongogogo: All list_*() are paginated. Need to handle more than 1 page. Either use a loop or the paginate() in boto

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,405 @@
...
+    ]
+    # Get the policies from the users groups
+    groups = [
+        group['GroupName'] for group in client.list_groups_for_user(UserName=username)['Groups']
```
jongogogo: All list_*() are paginated. Need to handle more than 1 page. Either use a loop or the paginate() in boto

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,405 @@
...
+    ]
+    for group in groups:
+        policies.extend([
+            policy['PolicyName'] for policy in client.list_attached_group_policies(GroupName=group)['AttachedPolicies']
```
jongogogo: All list_*() are paginated. Need to handle more than 1 page. Either use a loop or the paginate() in boto

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,405 @@
...
+        ])
+
+    return policies
+
+
+def get_role_policies(role_name):
+    client = boto3.client('iam')
+    # Get the users managed policies
+    return [
+        policy['PolicyName'] for policy in client.list_attached_role_policies(RoleName=role_name)['AttachedPolicies']
```
jongogogo: All list_*() are paginated. Need to handle more than 1 page. Either use a loop or the paginate() in boto

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,405 @@
...
+    for group in groups:
+        policies.extend([
+            policy['PolicyName'] for policy in client.list_attached_group_policies(GroupName=group)['AttachedPolicies']
+        ])
+
+    return policies
+
+
+def get_role_policies(role_name):
+    client = boto3.client('iam')
```
jongogogo: Use get_client() function to enable this rule to be used in a cross-account architecture

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,405 @@
...
+
+
+def should_ignore_config_item(config_item, ignored_roles, ignored_users):
+    return (config_item['configurationItemStatus'] == 'ResourceDeleted')\
+            or (config_item['resourceType'] == 'AWS::IAM::Role' and config_item['resourceName'] in ignored_roles)\
+            or (config_item['resourceType'] == 'AWS::IAM::User' and config_item['resourceName'] in ignored_users)\
+            or (config_item['ARN'].rsplit("/")[1] == 'aws-service-role')
+
+
+def get_user_policies(username):
```
jongogogo: Every API call is expensive for the service, please check iteratively if the policy is attach to the user, then if not the first group, then if not the second group etc. 
The goal is to reduce the number of call outside Config. Why not using the Configuration_item info, by the way?

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,405 @@
...
+    1 -- if a resource is deleted and generate a configuration change with ResourceDeleted status, the Boilerplate code will put a NOT_APPLICABLE on this resource automatically.
+    2 -- if a None or a list of dictionary is returned, the old evaluation(s) which are not returned in the new evaluation list are returned as NOT_APPLICABLE by the Boilerplate code
+    3 -- if None or an empty string, list or dict is returned, the Boilerplate code will put a "shadow" evaluation to feedback that the evaluation took place properly
+    """
+    ignored_roles = valid_rule_parameters["ignored_roles"]
+    ignored_users = valid_rule_parameters["ignored_users"]
+    policy_name = valid_rule_parameters['policy_name']
+
+    if should_ignore_config_item(configuration_item, ignored_roles, ignored_users):
+        return 'NOT_APPLICABLE'
```
jongogogo: I'd suggest to return COMPLIANT with an annotation on the ignored_roles and ignored_users

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,405 @@
...
+    configuration_item -- the configurationItem dictionary in the invokingEvent
+    valid_rule_parameters -- the output of the evaluate_parameters() representing validated parameters of the Config Rule
+
+    Advanced Notes:
+    1 -- if a resource is deleted and generate a configuration change with ResourceDeleted status, the Boilerplate code will put a NOT_APPLICABLE on this resource automatically.
+    2 -- if a None or a list of dictionary is returned, the old evaluation(s) which are not returned in the new evaluation list are returned as NOT_APPLICABLE by the Boilerplate code
+    3 -- if None or an empty string, list or dict is returned, the Boilerplate code will put a "shadow" evaluation to feedback that the evaluation took place properly
+    """
+    ignored_roles = valid_rule_parameters["ignored_roles"]
+    ignored_users = valid_rule_parameters["ignored_users"]
```
jongogogo: The name of the parameters are CamelCase by convention with the service team.

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,405 @@
...
+def evaluate_parameters(rule_parameters):
+    """Evaluate the rule parameters dictionary validity. Raise a ValueError for invalid parameters.
+
+    Return:
+    anything suitable for the evaluate_compliance()
+
+    Keyword arguments:
+    rule_parameters -- the Key/Value dictionary of the Config Rules parameters
+    """
+    valid_rule_parameters = rule_parameters
```
jongogogo: I'd suggest to use the ExceptionList the same way the service team is doing: https://docs.aws.amazon.com/config/latest/developerguide/iam-policy-blacklisted-check.html

python/IAM_POLICY_REQUIRED/parameters.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Version": "1.0", 
+  "Parameters": {
+    "CodeKey": "IAM_POLICY_REQUIRED.zip", 
+    "SourceRuntime": "python3.6", 
+    "RuleName": "IAM_POLICY_REQUIRED",
+    "SourceEvents": "AWS::IAM::Role,AWS::IAM::User", 
+    "OptionalParameters": "{}", 
+    "InputParameters": "{\"policy_name\": \"\", \"ignored_users\": \"\", \"ignored_roles\": \"\"}"
```
jongogogo: IgnoredUsers and IgnoredRoles are optional parameters, if I understand well.

HieronymusLex:
thanks for the feedback, I'll get these updates made tomorrow

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
HieronymusLex: good call - I wonder if we should use a list of policyArns instead of just 1 policy for consistency with the IAM Policy Blacklisted Check?

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
HieronymusLex: semantically it feels like NOT_APPLICABLE is more appropriate, since we're saying the resources should be excluded from evaluation - they weren't evaluated so they're neither COMPLIANT nor NON_COMPLIANT. However, I checked the IAM policy blacklisted and that returns COMPLIANT so I'll go with that for consistency

HieronymusLex:
I think a `CONTRIBUTING.md` would make help to capture some of this stuff like use of `get_client`, parameter naming conventions

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
HieronymusLex: fair point, will switch to using the configuration item info where possible

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
HieronymusLex: API call will no longer be used as switching to config item info

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
HieronymusLex: pagination feels like a bit of a premature optimisation here given the service limits for groups but just for future proof purposes I can add this in

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
HieronymusLex: API call no longer used

HieronymusLex: pagination feels like a bit of a premature optimisation here given the service limits for groups but just for future proof purposes I can add this in

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
HieronymusLex: API call no longer used

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
jongogogo: Good point, I resolve that

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
jongogogo: Good point

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
jongogogo: +1

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,403 @@
...
+    ignored_users = valid_rule_parameters["ignoredUsers"]
+    policy_name = valid_rule_parameters['policyName']
+    client = get_client('iam', event)
+
+    if should_ignore_config_item(configuration_item, ignored_roles, ignored_users):
+        return 'COMPLIANT'
+    elif has_policy_attached(configuration_item['resourceName'], configuration_item['resourceType'], policy_name, client):
+        return 'COMPLIANT'
+    else:
+        return 'NON_COMPLIANT'
```
jongogogo: Add annotation with the helper build_evaluation_from_config_item()

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,403 @@
...
+    2 -- if a None or a list of dictionary is returned, the old evaluation(s) which are not returned in the new evaluation list are returned as NOT_APPLICABLE by the Boilerplate code
+    3 -- if None or an empty string, list or dict is returned, the Boilerplate code will put a "shadow" evaluation to feedback that the evaluation took place properly
+    """
+    ignored_roles = valid_rule_parameters["ignoredRoles"]
+    ignored_users = valid_rule_parameters["ignoredUsers"]
+    policy_name = valid_rule_parameters['policyName']
+    client = get_client('iam', event)
+
+    if should_ignore_config_item(configuration_item, ignored_roles, ignored_users):
+        return 'COMPLIANT'
```
jongogogo: Add annotation with the helper build_evaluation_from_config_item()

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,403 @@
...
+    return (config_item['resourceType'] == 'AWS::IAM::Role' and config_item['resourceName'] in ignored_roles) \
+            or (config_item['resourceType'] == 'AWS::IAM::User' and config_item['resourceName'] in ignored_users) \
+            or (config_item['ARN'].rsplit("/")[1] == 'aws-service-role')
+
+
+def get_user_policies(username, client):
+    # Get the users managed policies
+    policies = [
+        policy['PolicyName'] for policy in client.list_attached_user_policies(UserName=username)['AttachedPolicies']
+    ]
```
jongogogo: Can you split this function in 2? So we can reduce by 1 API call if the policy is attach to the user.

HieronymusLex: yeah I'm going to refactor this code anyways to paginate the listing of groups, just in case the service limits change drastically

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
HieronymusLex: is there an example of dealing with comma separated lists (does it come in as a list or a string I have to split) and parameters structured like the ExceptionList for the example you gave i.e. `users:[user1, user2],roles:[role1, role2]`? Just want to make sure i'm not reinventing the wheel for parsing the args

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,405 @@
...
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may
+# not use this file except in compliance with the License. A copy of the License is located at
+#
+#        http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for
+# the specific language governing permissions and limitations under the License.
+
```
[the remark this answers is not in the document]
HieronymusLex: added

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
HieronymusLex: fixed

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
HieronymusLex: added

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
HieronymusLex: added

HieronymusLex:
@jongogogo I've done a fair bit of refactoring to bring this in line with the standards, use pagination, use the config item. one thing i noticed whilst testing is the RDK doesn't appear to be respecting the `OptionalParameters` in `parameters.json` when deploying the rule i.e. it gets deployed without it even when a value is set?

jongogogo:
> 
> 
> @jongogogo I've done a fair bit of refactoring to bring this in line with the standards, use pagination, use the config item. let me know if other changes are required.
> 
> one thing i noticed whilst testing is the RDK doesn't appear to be respecting the `OptionalParameters` in `parameters.json` when deploying the rule i.e. it gets deployed without it even when a value is set?

@HieronymusLex Yep, you are right. Do you mind +1 on this issue to prioritize the rdk team: https://github.com/awslabs/aws-config-rdk/issues/108

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,526 @@
...
+    pattern = re.compile("arn:(aws[a-zA-Z-]*)?:iam::(aws|\d{12}):policy\/[a-zA-Z0-9-_\/]+")
+    return pattern.match(arn)
+
+
+def extract_entities_from_exception_list(entity_type, exception_list):
+    pattern = re.compile("{Type}:\s?\[([a-zA-Z0-9-_,]+)\]".format(Type=entity_type))
+    matches = pattern.search(exception_list)
+    if matches:
+        return matches.group(1).replace(' ', '').split(",")
+    else:
```
jongogogo: Remove this `else:`, as the if statement include a return.

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,526 @@
...
+   To check IAM users and roles have a given policy attached directly or through a group.
+ Trigger:
+   Configuration Change on AWS::IAM::User/AWS::IAM::Role
+ Reports on:
+   AWS::IAM::User,AWS::IAM::Role
+ Rule Parameters:
+   | ---------------------- | --------- | -------------------------------------------------------- |
+   | Parameter Name         | Type      | Description                                              |
+   | ---------------------- | --------- | -------------------------------------------------------- |
+   | policyArns             | Required  | ARNs of the policies which should be attached to all     |
```
jongogogo: Please add an example in the description. It seems that you allow list and/or string. I'd settle on one or this other personally, no need to support both.

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,526 @@
...
+   |                        |           | users and roles.                                         |
+   | ---------------------- | --------- | -------------------------------------------------------- |
+   | exceptionList          | Optional  | Represents the IAM users and roles which are exempted    |
+   |                        |           | from the IAM Config rule. The valid users in this list   |
```
jongogogo: instead of `valid users`, put `valid entities` since there is users and roles.

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,526 @@
...
+   |                        |           | will be compliant by default.                            |
+   |                        |           | List of all the User Names separated by a comma          |
```
jongogogo: Same here, `entities`

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,526 @@
...
+ Reports on:
+   AWS::IAM::User,AWS::IAM::Role
+ Rule Parameters:
+   | ---------------------- | --------- | -------------------------------------------------------- |
+   | Parameter Name         | Type      | Description                                              |
+   | ---------------------- | --------- | -------------------------------------------------------- |
+   | policyArns             | Required  | ARNs of the policies which should be attached to all     |
+   |                        |           | users and roles.                                         |
+   | ---------------------- | --------- | -------------------------------------------------------- |
+   | exceptionList          | Optional  | Represents the IAM users and roles which are exempted    |
```
jongogogo: Is the code deal well with user/role having a path? 
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-friendly-names

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,526 @@
...
+   |                        |           | List of all the User Names separated by a comma          |
+   | ---------------------- | --------- | -------------------------------------------------------- |
+ Feature:
+   As: a Security Officer
+   I want: To ensure that IAM roles and users have mandatory policies attached
+   In order to: enforce mandatory permissions for IAM entities
+ Scenarios:
+  Scenario: 1
+     Given: No IAM Users or Roles exist
+      When: Evaluation occurs
```
jongogogo: I'd remove the "when" statement in all scenarios. While it is true, I'd never saw it requested by anyone to understand the gherkin.

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,526 @@
...
+      Then: Return "NOT_APPLICABLE"
+   Scenario: 2
+     Given: exceptionList is configured
+       And: A user listed in the exceptionList is not an alphanumerical string
+      When: Evaluation occurs
+      Then: Return an error
+   Scenario: 3
+     Given: exceptionList is configured
+       And: A role listed in the exceptionList is not an alphanumerical string
```
jongogogo: For scenario having just one term which change, you can use the scenario outline format: https://stackoverflow.com/questions/9093872/gherkin-or-syntax-to-reduce-repetition-with-bdd

```
   Scenario: 2
     Given: exceptionList is configured
       And: A <entity> listed in the exceptionList is not an alphanumerical string
      Then: Return an error

   Examples:
  | entity     |
  | IAM User   |
  | IAM Role   |
```

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,526 @@
...
+            attached_policies = paginate(client, client.list_attached_group_policies, **{'GroupName': group})
+            for policy in attached_policies:
+                managed_policies.append(policy['PolicyArn'])
+                if list_contains_all(managed_policies, policy_arns):
+                    return True
+
+        return False
+    elif resource_type == 'AWS::IAM::Role':
+        return list_contains_all(get_attached_policies(configuration_item), policy_arns)
+    else:
```
jongogogo: Remove `else:` since all the other statement `return`

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
```diff
@@ -0,0 +1,526 @@
...
+
+    Keyword arguments:
+    rule_parameters -- the Key/Value dictionary of the Config Rules parameters
+    """
+    policy_arns = rule_parameters["policyArns"]
+    parsed_policy_arns = policy_arns if isinstance(policy_arns, list) else policy_arns.split(",")
+    if not all(is_valid_arn(arn) for arn in parsed_policy_arns):
+        raise ValueError('Invalid policy ARNs specified in policyArns')
+
+    exception_list = rule_parameters.get("exceptionList", "")
```
jongogogo: Didn't know get(). Nice :)

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
HieronymusLex: no problem, will go with list

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
HieronymusLex: yeah the path has no effect since we only consider the name which comes in the event

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
HieronymusLex: updated

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED.py
jongogogo: Agree, just the WHEN statement is always the same in this case, hence the fact that it has no value here.

python/IAM_POLICY_REQUIRED/IAM_POLICY_REQUIRED_test.py
```diff
@@ -0,0 +1,363 @@
...
+        "messageType": "ConfigurationItemChangeNotification",
+        "recordVersion": "1.3"
+    }
+
+
+rule = __import__('IAM_POLICY_REQUIRED')
+
+class TestPolicyRequired(unittest.TestCase):
+
+    rule_parameters = '{"policyArns":["arn:aws:iam::aws:policy/AdministratorAccess"], "exceptionList": ""}'
```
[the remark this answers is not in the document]
HieronymusLex: the RDK boilerplate uses `rule_parameters = json.loads(event['ruleParameters'])` to load the rule parameters therefore this comes in as a list. Similarly, if you pass a list type in through CFN (e.g. a parameter of type comma separated list, it will be a Python List when passed to the rule without any string splitting required etc.

I could change it to a string, but it would mean parsing would need to be added in the Lambda function custom code rather than handled by the RDK boilerplate. what do you think?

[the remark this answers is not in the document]
jongogogo: ```event['ruleParameters']``` comes as a dir. The value of each key of this dir must be a string. Currently your value are list, so it should not be working if you try to deploy it in the Config service.

Incorrect:
```'{"policyArns":["arn:aws:iam::aws:policy/AdministratorAccess"], "exceptionList": ""}'```

Correct:
```'{"policyArns":"[arn:aws:iam::aws:policy/AdministratorAccess]", "exceptionList": ""}'```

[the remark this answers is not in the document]
HieronymusLex: @jongogogo updated!

PR #124, opened 2018-11-27
Config Rule to check whether your CloudFront Distribution has been co…

ppratee:
…nfigured to store logs on an authorized S3 bucket

Config Rule to check whether your CloudFront Distribution has been configured to store logs on an authorized S3 bucket

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/python/cf_logging_enabled/cf_logging_enabled.py
+++ b/python/cf_logging_enabled/cf_logging_enabled.py
@@ -91,11 +91,10 @@ def evaluate_compliance(event, configuration_item, valid_rule_parameters):
     logging_bucket = valid_rule_parameters ['CentralLoggingBucket'] + '.s3.amazonaws.com'
     if not configuration_item['configuration']['distributionConfig']['logging']['enabled']:
         return build_evaluation_from_config_item(configuration_item, 'NON_COMPLIANT', annotation='Distribution is not configured to store logs.')
-    else:
-        if configuration_item['configuration']['distributionConfig']['logging']['bucket'] == logging_bucket:
-            return build_evaluation_from_config_item(configuration_item, 'COMPLIANT')
-        else:
-            return build_evaluation_from_config_item(configuration_item, 'NON_COMPLIANT', annotation='Distribution is configured to store logs in an unauthorized bucket.')
+    if configuration_item['configuration']['distributionConfig']['logging']['bucket'] == logging_bucket:
+        return build_evaluation_from_config_item(configuration_item, 'COMPLIANT')
+    if not configuration_item['configuration']['distributionConfig']['logging']['bucket'] == logging_bucket:
+        return build_evaluation_from_config_item(configuration_item, 'NON_COMPLIANT', annotation='Distribution is configured to store logs in an unauthorized bucket.')
 
 def evaluate_parameters(rule_parameters):
     """Evaluate the rule parameters dictionary validity. Raise a ValueError for invalid parameters.
--- a/python/cf_logging_enabled/cf_logging_enabled.py
+++ b/python/cf_logging_enabled/cf_logging_enabled.py
@@ -90,12 +90,12 @@ def evaluate_compliance(event, configuration_item, valid_rule_parameters):
     
     logging_bucket = valid_rule_parameters ['CentralLoggingBucket'] + '.s3.amazonaws.com'
     if not configuration_item['configuration']['distributionConfig']['logging']['enabled']:
-        return build_evaluation_from_config_item(configuration_item, 'NON_COMPLIANT', annotation='distribution is not configured to store logs')
+        return build_evaluation_from_config_item(configuration_item, 'NON_COMPLIANT', annotation='Distribution is not configured to store logs.')
     else:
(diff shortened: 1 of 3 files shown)


python/cf_logging_enabled/cf_logging_enabled.py
```diff
@@ -0,0 +1,396 @@
...
+  Checks whether your CloudFront Distribution has been configured to store logs on an authorized S3 bucket
+Trigger:
+  Configuration change on AWS::CloudFront::Distribution
+Resource Type to report on:
+  AWS::CloudFront::Distribution
+Rule Parameters:
+  | ---------------------- | ---------- | -------------------------------------------------------------- |
+  | Parameter Name         | Type       | Description                                                    |
+  | ---------------------- | ---------- | -------------------------------------------------------------- |
+  | CentralLoggingBucket   | Mandatory  | Authorized S3 Bucket for storing CloudFront logs               |
```
jongogogo: Why not making this parameter optional? I would allow to just verify that logging is enabled.

python/cf_logging_enabled/cf_logging_enabled.py
```diff
@@ -0,0 +1,396 @@
...
+    """
+
+    ###############################
+    # Add your custom logic here. #
+    ###############################
+    
+    logging_bucket = valid_rule_parameters ['CentralLoggingBucket'] + '.s3.amazonaws.com'
+    if not configuration_item['configuration']['distributionConfig']['logging']['enabled']:
+        return build_evaluation_from_config_item(configuration_item, 'NON_COMPLIANT', annotation='distribution is not configured to store logs')
+    else:
```
jongogogo: This `else:` is not necessary, since the `if` statement is a `return`

python/cf_logging_enabled/cf_logging_enabled.py
```diff
@@ -0,0 +1,396 @@
...
+        if configuration_item['configuration']['distributionConfig']['logging']['bucket'] == logging_bucket:
+            return build_evaluation_from_config_item(configuration_item, 'COMPLIANT', annotation='distribution is configured to store logs in authorized bucket')
+        else:
```
jongogogo: idem

python/cf_logging_enabled/cf_logging_enabled.py
```diff
@@ -0,0 +1,396 @@
...
+    ###############################
+    # Add your custom logic here. #
+    ###############################
+    
+    logging_bucket = valid_rule_parameters ['CentralLoggingBucket'] + '.s3.amazonaws.com'
+    if not configuration_item['configuration']['distributionConfig']['logging']['enabled']:
+        return build_evaluation_from_config_item(configuration_item, 'NON_COMPLIANT', annotation='distribution is not configured to store logs')
+    else:
+        if configuration_item['configuration']['distributionConfig']['logging']['bucket'] == logging_bucket:
+            return build_evaluation_from_config_item(configuration_item, 'COMPLIANT', annotation='distribution is configured to store logs in authorized bucket')
```
jongogogo: No need of an annotation

python/cf_logging_enabled/cf_logging_enabled.py
```diff
@@ -0,0 +1,396 @@
...
+    3 -- if None or an empty string, list or dict is returned, the Boilerplate code will put a "shadow" evaluation to feedback that the evaluation took place properly
+    """
+
+    ###############################
+    # Add your custom logic here. #
+    ###############################
+    
+    logging_bucket = valid_rule_parameters ['CentralLoggingBucket'] + '.s3.amazonaws.com'
+    if not configuration_item['configuration']['distributionConfig']['logging']['enabled']:
+        return build_evaluation_from_config_item(configuration_item, 'NON_COMPLIANT', annotation='distribution is not configured to store logs')
```
jongogogo: Start annotation with a capital letter and ending by dot.

python/cf_logging_enabled/cf_logging_enabled.py
ppratee: I can make it optional. However, from the functionality perspective, if logs are not getting stored in proper bucket, then the monitoring / archival and other controls around logs might not get configured appropriately. As such, there would be no / very limited value of enabling the logging if the logs are not propoerly stored / monitored / archived. Hence, I think it should be a mandatory parameter. However, if it is a standard to make parameters optional, I am happy to do so.

python/cf_logging_enabled/cf_logging_enabled.py
ppratee: Updated the code

python/cf_logging_enabled/cf_logging_enabled.py
ppratee: updated the code

ppratee: updated the code

python/cf_logging_enabled/cf_logging_enabled.py
ppratee: updated the code

ppratee:
@jongogogo - Any updates?

PR #174, opened 2019-04-01
Initial Commit:  EC2_ENCRYPTED_VOLUMES

dewjam:

*Issue #, if available:*
N/A

*Description of changes:*
Adds a new rule that checks EC2 instances to ensure all volumes are encrypted.  The difference between this rule and `EBS_ENCRYPTED_VOLUMES_V2` is that this rule triggers on EC2 Instance configuration changes rather than EBS volume configuration changes.  For some background, there are some situations where EBS volume CIs are not updated in realtime and rule evaluation becomes heavily delayed.  By triggering on EC2 Instances, we work around this problem and provide very similar functionality to `EBS_ENCRYPTED_VOLUMES_V2`.


jongogogo:
As discuss please resubmit a commit to re-run the check (it is working now). Thanks!

PR #220, opened 2019-04-26
Adding new rule: CLOUDTRAIL_S3_DATAEVENTS_ENABLED

Kartik107:

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
+++ b/python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
@@ -106,7 +106,10 @@ def evaluate_compliance(event, configuration_item, valid_rule_parameters):
     if not valid_rule_parameters:
         for trail in trail_list:
             if trail['HasCustomEventSelectors']:
-                event_selector_output = ct_client.get_event_selectors(TrailName=trail['Name'])['EventSelectors']
+                #example arn: arn:aws:cloudtrail:us-east-1:123456789012:trail/test
+                region = trail['TrailARN'][19:].split(':')[0]
+                ct_regional_client = boto3.client('cloudtrail', region_name=region)                
+                event_selector_output = ct_regional_client.get_event_selectors(TrailName=trail['Name'])['EventSelectors']
                 data_resource_values = get_data_resource_s3_values(event_selector_output)
                 if isinstance(data_resource_values, bool):
                     evaluations.append(build_evaluation(event['accountId'], 'COMPLIANT', event, DEFAULT_RESOURCE_TYPE))
@@ -116,7 +119,10 @@ def evaluate_compliance(event, configuration_item, valid_rule_parameters):
     else:
         for trail in trail_list:
             if trail['HasCustomEventSelectors']:
-                event_selector_output = ct_client.get_event_selectors(TrailName=trail['Name'])['EventSelectors']
+                #example arn: arn:aws:cloudtrail:us-east-1:123456789012:trail/test
+                region = trail['TrailARN'][19:].split(':')[0]
+                ct_regional_client = boto3.client('cloudtrail', region_name=region)                
+                event_selector_output = ct_regional_client.get_event_selectors(TrailName=trail['Name'])['EventSelectors']
                 data_resource_values = get_data_resource_s3_values(event_selector_output)
                 #if logging is enabled for 'arn:aws:s3', return COMPLIANT
                 if isinstance(data_resource_values, bool):
--- a/python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
+++ b/python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
@@ -8,7 +8,7 @@
 # or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for
 # the specific language governing permissions and limitations under the License.
-'''
+"""
 Description:
   Checks that at least one AWS CloudTrail trail logs data events for the Amazon S3 service buckets. The rule returns NON_COMPLIANT if none of the trails log data events for Amazon S3 service buckets. The rule checks for trails in their HomeRegion only. The rule does not support S3 prefixes.
 Trigger:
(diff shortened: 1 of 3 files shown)


python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
```diff
@@ -0,0 +1,489 @@
...
+def evaluate_compliance(event, configuration_item, valid_rule_parameters):
+    ct_client = get_client('cloudtrail', event)
+    evaluations = []
+    compliant_buckets = []
+    non_compliant_buckets = []
+    trail_list = get_all_trails(ct_client)
+
+    #No trails configured: return NON_COMPLIANT
+    if not trail_list:
+        evaluations.append(build_evaluation(event['accountId'], 'NON_COMPLIANT', event, DEFAULT_RESOURCE_TYPE, 'There are no AWS CloudTrail Trails in the region.'))
```
shikharj05: We can change this line to 
evaluations.append(build_evaluation(event['accountId'], 'NON_COMPLIANT', event, annotation='There are no AWS CloudTrail Trails in the region.'))

the build evaluation function already has a default value for resource_type=DEFAULT_RESOURCE_TYPE

python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
```diff
@@ -0,0 +1,489 @@
...
+        return evaluations
+
+    #No rule parameter specified, default check for all S3 buckets
+    if not valid_rule_parameters:
+        for trail in trail_list:
+            if trail['HasCustomEventSelectors']:
+                #example arn: arn:aws:cloudtrail:us-east-1:123456789012:trail/test
+                region = trail['TrailARN'][19:].split(':')[0]
+                ct_regional_client = boto3.client('cloudtrail', region_name=region)
```
shikharj05: Add a comment here to explain why we need a regional client

python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
shikharj05: We need to change the annotation here: 'There are no AWS CloudTrail Trails configured to log events in this region.'

python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
```diff
@@ -0,0 +1,489 @@
...
+                    compliant_buckets.append(bucket_name)
+
+        #match rule parameter values v/s buckets with data events logging enabled
+        non_compliant_buckets = (list(set(valid_rule_parameters) - set(compliant_buckets)))
+        #If non_compliant_buckets is empty, logging is enabled for all required buckets
+        if not non_compliant_buckets:
+            evaluations.append(build_evaluation(event['accountId'], 'COMPLIANT', event, DEFAULT_RESOURCE_TYPE))
+            return evaluations
+
+        evaluations.append(build_evaluation(event['accountId'], 'NON_COMPLIANT', event, DEFAULT_RESOURCE_TYPE, 'Not valid for some buckets: ' + str(non_compliant_buckets) + '.'))
```
shikharj05: Change annotation to-'AWS CloudTrail trails do not log S3 data events for buckets: ' + str(non_compliant_buckets) + '.'

python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
```diff
@@ -96,38 +96,40 @@ def evaluate_compliance(event, configuration_item, valid_rule_parameters):
...
-                #if logging is enabled for 'arn:aws:s3', return COMPLIANT
-                if isinstance(data_resource_values, bool):
-                    evaluations.append(build_evaluation(event['accountId'], 'COMPLIANT', event, DEFAULT_RESOURCE_TYPE))
-                    return evaluations
+            trail_region = trail['HomeRegion']
+            if trail_region == region:
+                if trail['HasCustomEventSelectors']:
+                    #example arn: arn:aws:cloudtrail:us-east-1:123456789012:trail/test
+                    region = trail['TrailARN'][19:].split(':')[0]
+                    ct_client = boto3.client('cloudtrail', region_name=region)
```
shikharj05: We do not need new client here, you can use the existing one

shikharj05:
Thank you for the changes, @jongogogo kindly review this PR & let us know if any changes are required

python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
```diff
@@ -0,0 +1,488 @@
...
+    return evaluations
+
+def get_all_trails(ct_client):
+    all_trails = []
+
+    trails_list = ct_client.describe_trails()
+    while True:
+        all_trails += trails_list['trailList']
+        if 'NextToken' in trails_list:
+            trails_list = ct_client.describe_trails(NextToken=trails_list['NextToken'])
```
jongogogo: Are you sure it works? I don't see the "NextToken" documented in boto: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudtrail.html#CloudTrail.Client.describe_trails

python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
```diff
@@ -0,0 +1,488 @@
...
+                    event_selector_output = ct_client.get_event_selectors(TrailName=trail['Name'])['EventSelectors']
+                    data_resource_values = get_data_resource_s3_values(event_selector_output)
+                    if isinstance(data_resource_values, bool):
+                        evaluations.append(build_evaluation(event['accountId'], 'COMPLIANT', event))
+                        return evaluations
+
+    #Rule parameter S3BucketName is specified
+    else:
+        for trail in trail_list:
+            trail_region = trail['HomeRegion']
```
jongogogo: We might have been on the wrong path here (apologies for that). It might be confusing for the customer to consider s3 buckets as parameters in different regions. Especially if they have a multi-region trail on all s3 buckets.

I see that we can filter shadow trails on-demand. Having an input parameter IncludeShadowTrail (optional) with a default to true, might be easier. 

You can then describe the appropriate trails on-demand with a filter on the DescribeTrails API call and not anymore care about the HomeRegion.

What do you think?

 https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_DescribeTrails.html

python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
```diff
@@ -0,0 +1,488 @@
...
+                    data_resource_values = get_data_resource_s3_values(event_selector_output)
+                    #if logging is enabled for 'arn:aws:s3', return COMPLIANT
+                    if isinstance(data_resource_values, bool):
+                        evaluations.append(build_evaluation(event['accountId'], 'COMPLIANT', event))
+                        return evaluations
+
+                #get bucket names from data resource values to store them in a list
+                for value in data_resource_values:
+                    # value is of the form arn:aws:s3:::test/, [13:-1] strips 'arn:aws:s3:::' and the trailing slash
+                    bucket_name = value[13:-1]
```
jongogogo: This is going to be challenging for other partitions (china, gov-cloud, etc.).

python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
```diff
@@ -0,0 +1,488 @@
...
+    #no rule parameter
+    if not rule_parameters:
+        return valid_rule_parameters
+
+    #invalid rule parameter
+    if 'S3BucketName' not in rule_parameters:
+        raise ValueError("S3BucketName is the only supported rule parameter")
+
+    bucket_names = [s3bucketname.strip() for s3bucketname in rule_parameters['S3BucketName'].split(',')]
+    valid_name_regex = re.compile("(?=^.{3,63}$)(?!^(\\d+\\.)+\\d+$)(^(([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])\\.)*([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])$)")
```
jongogogo: Nice :)

python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
shikharj05: Agreed, will try to split using ':' and extract value using the split.

python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
shikharj05: Not sure if this will affect the code logic as IncludeShadowTrail is set to True by default, we get all trails and need to check for HomeRegion. 
If set to False, we won't get other region trails but if we add this as a parameter, we will have to check for it's value and then decide whether to check for HomeRegion or not.

Basically, we would be removing one check and adding another.

We can start as default of False, this will limit the number of trails returned and reduce the output of API call.

python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
shikharj05: We re-used the code from here: https://github.com/awslabs/aws-config-rules/blob/master/python/CLOUDTRAIL_ENABLED_V2/CLOUDTRAIL_ENABLED_V2.py#L236

Since the rule was in the repo, we didn't use additional checks to test this. Checking using CLI, I don't think this would work as this API doesn't support --max-items as well.
```
aws cloudtrail describe-trails --max-items 1
usage: aws [options] <command> <subcommand> [<subcommand> ...] [parameters]
To see help text, you can run:

  aws help
  aws <command> help
  aws <command> <subcommand> help

Unknown options: --max-items, 1
```
We can remove this.

jongogogo:
Any actions needed from me there?

shikharj05:
@jongogogo Nothing from your end, will put the new PR up soon.

jongogogo:
When the update is planned?

python/CLOUDTRAIL_S3_DATAEVENTS_ENABLED/CLOUDTRAIL_S3_DATAEVENTS_ENABLED.py
jongogogo: Need update?

Issue #199, opened 2019-04-17
New Rule GUARDDUTY_UNTREATED_FINDINGS


rafihm:
```
#####################################
##           Gherkin               ##
#####################################
Rule Name:
    GUARDDUTY_UNTREATED_FINDINGS

Description:
    Checks whether the GuardDuty has untreated findings. The rule is NON_COMPLIANT if the GuardDuty has untreated finding older than X days.

Trigger:
    Periodic

Reports on:
    AWS::::Account

Rule Parameters:
  daysLowSev (Optional)
    The number of days the GuardDurty Low severity findings are allowed to stay untreated (default: 30 days)

  daysMediumSev (Optional)
    The number of days the GuardDurty Medium severity findings are allowed to stay untreated (default: 7 days)

  daysHighSev (Optional)
    The number of days the GuardDurty High severity findings are allowed to stay untreated (default: 1 day)

Scenarios:
  Scenario: 1
    Given: GuardDuty is Disabled or Suspended.
    Then: Return NOT_APPLICABLE

  Scenario: 2
    Given: The rule parameters daysLowSev/daysMediumSev/daysHighSev has invalid value.
    Then: Return ERROR

  Scenario: 3
    Given: No untreated GuardDuty findings are present.
    Then: Return COMPLIANT

  Scenario: 4
    Given: GuardDuty has untreated Low/Medium/High Severity findings.
    And: Low/Medium/High Severity findings are older than daysLowSev/daysMediumSev/daysHighSev number of days respectively.
    Then: Return NON_COMPLIANT

  Scenario: 5
    Given: GuardDuty has untreated Low/Medium/High Severity findings.
    And: Low/Medium/High Severity findings are not older than daysLowSev/daysMediumSev/daysHighSev number of days respectively.
    Then: Return COMPLIANT
```

**Required API calls**
[ListDetectors](https://docs.aws.amazon.com/guardduty/latest/ug/list-detectors.html) - if it return empty list then guard duty is disabled.

[GetDetector](https://docs.aws.amazon.com/guardduty/latest/ug/get-detector.html) - Takes detector ID as parameter, this API call used to check whether GuardDuty is suspended, if it is [suspended](https://docs.aws.amazon.com/guardduty/latest/ug/guardduty_suspend-disable.html) then it returns `"Status": "DISABLED"`
if the GuardDuty is suspended then it will still show all the existing finding but those findings will not get updated.

[...]

PR #231, opened 2019-05-01
adding new rule GUARDDUTY_UNTREATED_FINDINGS

rafihm:

*Issue #, if available:*
#199 
*Description of changes:*
Initial commit for new rule GUARDDUTY_UNTREATED_FINDINGS


python/GUARDDUTY_UNTREATED_FINDINGS/GUARDDUTY_UNTREATED_FINDINGS.py
```diff
@@ -0,0 +1,509 @@
...
+
+
+def evaluate_parameters(rule_parameters):
+    default_days_low_sev = 30
+    default_days_medium_sev = 7
+    default_days_high_sev = 1
+
+    if "daysLowSev" not in rule_parameters:
+        rule_parameters['daysLowSev'] = default_days_low_sev
+    elif not validate_parameter_value(rule_parameters['daysLowSev']):
```
5u5hma: I think this should be else or if not. I think pylint would throw unnecessary elif.

python/GUARDDUTY_UNTREATED_FINDINGS/GUARDDUTY_UNTREATED_FINDINGS.py
```diff
@@ -0,0 +1,509 @@
...
+    if "daysHighSev" not in rule_parameters:
+        rule_parameters['daysHighSev'] = default_days_high_sev
+    elif not validate_parameter_value(rule_parameters['daysHighSev']):
+        rule_parameters['daysHighSev'] = default_days_high_sev
+
+    #change string values to int
+    rule_parameters['daysLowSev'] = int(float(rule_parameters['daysLowSev']))
+    rule_parameters['daysMediumSev'] = int(float(rule_parameters['daysMediumSev']))
+    rule_parameters['daysHighSev'] = int(float(rule_parameters['daysHighSev']))
+
```
5u5hma: Are we considering decimals for the days value? 
If it is integer value then we can use the int(value) directly. We can check in the parameters to ensure we do not have decimal values in days similar to that of positive value checking. Please do correct me if I am thinking in wrong direction.

python/GUARDDUTY_UNTREATED_FINDINGS/GUARDDUTY_UNTREATED_FINDINGS.py
```diff
@@ -0,0 +1,509 @@
...
+    return rule_parameters
+
+# returns False if the given value is not a Positive Number.
+def validate_parameter_value(parameter_value):
+    try:
+        float(parameter_value)
```
5u5hma: If the above said is considered true then we can use int(value) here as well.

python/GUARDDUTY_UNTREATED_FINDINGS/GUARDDUTY_UNTREATED_FINDINGS.py
```diff
@@ -0,0 +1,509 @@
...
+    creation_datetime_object = datetime.strptime(finding_creation_time, '%Y-%m-%dT%H:%M:%S.%fZ')
+    delta_time = current_datetime_object - creation_datetime_object
+    return delta_time.days
+
+# function to get all findings. Each call supports maximum of 50 items.
+def list_all_findings(guardduty_client, guardduty_detector_id):
+    all_findings_list = []
+    findings_id_list = guardduty_client.list_findings(DetectorId=guardduty_detector_id)
+    if not findings_id_list['FindingIds']:
+        return []
```
5u5hma: Can we return all_findings_list here as it will be a empty list initially and it will be equivalent to returning []

python/GUARDDUTY_UNTREATED_FINDINGS/GUARDDUTY_UNTREATED_FINDINGS.py
```diff
@@ -0,0 +1,509 @@
...
+    # evaluate each finding
+    for each_finding in all_findings_list:
+        days_since_created = get_delta_days(each_finding['CreatedAt'])
+
+        # 0.1 to 3.9 severity score is LowSev
+        if 0.1 <= each_finding['Severity'] <= 3.9:
+            if days_since_created <= valid_rule_parameters['daysLowSev']:
+                evaluations.append(build_evaluation(each_finding['Id'], 'COMPLIANT', event))
+                continue
+            evaluations.append(build_evaluation(each_finding['Id'], 'NON_COMPLIANT', event, annotation='This AWS GurdDuty Low Severity finding is older than {} days'.format(valid_rule_parameters['daysLowSev'])))
```
5u5hma: Please add the final dot after "days" word in both the annotation.

python/GUARDDUTY_UNTREATED_FINDINGS/GUARDDUTY_UNTREATED_FINDINGS.py
rafihm: I dont think we can use `else` here as we need to validate the parameter and make the decision. Let me know if missed something here.

python/GUARDDUTY_UNTREATED_FINDINGS/GUARDDUTY_UNTREATED_FINDINGS.py
rafihm: The parameters passed will be in string data type. If the user is passed value with the decimals (ex: 10.0 ) then we need to treat it as a valid parameter. Treating this as invalid parameter generating error causes confusion as it is basically valid input.
I have used to `float->int` conversion as you cannot convert the `string` to `int` directly if it has decimal.

python/GUARDDUTY_UNTREATED_FINDINGS/GUARDDUTY_UNTREATED_FINDINGS.py
rafihm: Modified it to return `all_findings_list`

python/GUARDDUTY_UNTREATED_FINDINGS/GUARDDUTY_UNTREATED_FINDINGS.py
rafihm: Done.

python/GUARDDUTY_UNTREATED_FINDINGS/GUARDDUTY_UNTREATED_FINDINGS.py
rafihm: Added `ValueError` if input is invalid.

rafihm:
@5u5hma Thanks for the review :) please check the comment and let me know if your thoughts about it.

5u5hma:
Thanks Rafih. @jongogogo Please review it and let us know for any changes.

python/GUARDDUTY_UNTREATED_FINDINGS/GUARDDUTY_UNTREATED_FINDINGS.py
5u5hma: @jongogogo  Please let us know your thoughts on this. If we restrict the user from provide decimal for days value we can avoid one level of conversion and directly convert the string to int. So let us know what do you think we need to go with here.

Issue #204, opened 2019-04-20
New rule for ALB HTTP to HTTPS redirection check


santosh07bec:
```python
'''
Rule Name:
  ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK

Description:
  Checks whether HTTP to HTTPS redirection is configured on all HTTP listeners of Application Load Balancers. The rule in NON_COMPLIANT if one or more HTTP listener of an Application Load Balancer do not have HTTP to HTTPS redirection configured.

Trigger:
  Periodic

Reports on:
  AWS::ElasticLoadBalancingV2::LoadBalancer

Rule Parameters:
  None

Scenarios:
  Scenario: 1
     Given: No Application Load Balancer is present
      Then: Return NOT_APPLICABLE
  Scenario: 2
     Given: There is at least 1 Application Load Balancer
       And: Application Load Balancer has only HTTPS listener(s)
      Then: Return COMPLIANT
  Scenario: 3
     Given: There is at least 1 Application Load Balancer
       And: Application Load Balancer has one or more HTTP listeners configured
       And: At least one HTTP listener rule does not have HTTP to HTTPS redirection action configured
      Then: Return NON_COMPLIANT
  Scenario: 4
     Given: There is at least 1 Application Load Balancer
       And: Application Load Balancer has one or more HTTP listeners configured
       And: All HTTP listener rules have HTTP to HTTPS redirection action configured
      Then: Return COMPLIANT
'''
```

jongogogo:
Updated, good to go

schirag1993:
Hello @santosh07bec 

I believe there is one more scenario:
Given: There is at least 1 Application Load Balancer
And: Application Load Balancer has no listeners configured

Does this scenario make sense?

jongogogo:
Interesting. +1 on new scenario, I'd suggest to return NOT_APPLICABLE.

PR #236, opened 2019-05-03
Rule :: ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK

ketanbhut:

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/python/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK.py
+++ b/python/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK.py
@@ -60,7 +60,6 @@ except ImportError:
 
 # Define the default resource to report to Config Rules
 DEFAULT_RESOURCE_TYPE = 'AWS::ElasticLoadBalancingV2::LoadBalancer'
-ACCOUNT_RESOURCE_TYPE = 'AWS::::Account'
 
 # Set to True to get the lambda to assume the Role attached on the Config Service (useful for cross-account).
 ASSUME_ROLE_MODE = False
@@ -77,7 +76,7 @@ def evaluate_compliance(event, configuration_item, valid_rule_parameters):
     alb_client = get_client("elbv2", event)
     all_elbv2 = get_all_elbv2(alb_client)
     if not all_elbv2:
-        return [build_evaluation(event['accountId'], 'NOT_APPLICABLE', event, resource_type=ACCOUNT_RESOURCE_TYPE)]
+        return [build_evaluation(event['accountId'], 'NOT_APPLICABLE', event, resource_type='AWS::::Account')]
     for elb in all_elbv2:
         if elb['Type'] != 'application':
             evaluations.append(build_evaluation(elb['LoadBalancerArn'], 'NOT_APPLICABLE', event))
@@ -85,14 +84,15 @@ def evaluate_compliance(event, configuration_item, valid_rule_parameters):
         alb_all_listeners = get_all_listeners(alb_client, elb['LoadBalancerArn'])
         overall_listeners_eval = 'NON_COMPLIANT'
         for lis in alb_all_listeners:
-            if not is_https_listener(lis):
-                for action in lis['DefaultActions']:
-                    if(action['Type'] == 'redirect' and action['RedirectConfig']['Protocol'] == 'HTTPS'):
-                        overall_listeners_eval = 'COMPLIANT'
-                        continue
-                    overall_listeners_eval = 'NON_COMPLIANT'
-            else:
+            if is_https_listener(lis):
                 overall_listeners_eval = 'COMPLIANT'
+                continue
+            for action in lis['DefaultActions']:
+                if action['Type'] == 'redirect' and action['RedirectConfig']['Protocol'] == 'HTTPS':
+                    overall_listeners_eval = 'COMPLIANT'
+                    continue
+                overall_listeners_eval = 'NON_COMPLIANT'
+                break
             if overall_listeners_eval == 'NON_COMPLIANT':
                 break
         evaluations.append(build_evaluation(elb['LoadBalancerArn'], overall_listeners_eval, event, annotation=get_str(overall_listeners_eval)))
--- a/python/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK.py
+++ b/python/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK.py
@@ -85,12 +85,16 @@ def evaluate_compliance(event, configuration_item, valid_rule_parameters):
             if is_https_listener(lis):
                 overall_listeners_eval = 'COMPLIANT'
                 continue
-            for action in lis['DefaultActions']:
-                if action['Type'] == 'redirect' and action['RedirectConfig']['Protocol'] == 'HTTPS':
-                    overall_listeners_eval = 'COMPLIANT'
-                    continue
-                overall_listeners_eval = 'NON_COMPLIANT'
-                break
+            rules = get_all_listener_rules(alb_client, lis['ListenerArn'])
+            for rule in rules:
+                for action in rule['Actions']:
+                    if action['Type'] == 'redirect' and action['RedirectConfig']['Protocol'] == 'HTTPS':
+                        overall_listeners_eval = 'COMPLIANT'
+                        continue
+                    overall_listeners_eval = 'NON_COMPLIANT'
+                    break
+                if overall_listeners_eval == 'NON_COMPLIANT':
+                    break
             if overall_listeners_eval == 'NON_COMPLIANT':
                 break
         evaluations.append(build_evaluation(elb['LoadBalancerArn'], overall_listeners_eval, event, annotation=get_str(overall_listeners_eval)))
@@ -120,6 +124,14 @@ def get_all_listeners(client, elbv2_arn):
     while resp:
         items += resp['Listeners']
         resp = client.describe_listeners(LoadBalancerArn=elbv2_arn, Marker=resp['NextMarker']) if 'NextMarker' in resp else None
+    return items
+
+def get_all_listener_rules(client, listener_arn):
+    resp = client.describe_rules(ListenerArn=listener_arn, PageSize=400)
+    items = []
+    while resp:
+        items += resp['Rules']
+        resp = client.describe_rules(ListenerArn=listener_arn, Marker=resp['NextMarker']) if 'NextMarker' in resp else None
     return items
 
 def evaluate_parameters(rule_parameters):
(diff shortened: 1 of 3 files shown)


python/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK.py
```diff
@@ -0,0 +1,436 @@
...
+ASSUME_ROLE_MODE = False
+
+# Other parameters (no change needed)
+CONFIG_ROLE_TIMEOUT_SECONDS = 900
+
+#############
+# Main Code #
+#############
+
+def evaluate_compliance(event, configuration_item, valid_rule_parameters):
```
schirag1993: Please remove all whitespaces present in this function @ketanbhut

python/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK.py
```diff
@@ -0,0 +1,436 @@
...
+        evaluations.append(build_evaluation(elb['LoadBalancerArn'], overall_listeners_eval, event, annotation=get_str(overall_listeners_eval)))
+    return evaluations
+
+def get_str(compliance_value):
+    if compliance_value == 'NON_COMPLIANT':
+        return "HTTP listener rule must have HTTP to HTTPS redirection action configured"
+    return None
+
+def is_https_listener(listener):
+    if 'SslPolicy' in listener.keys():
```
schirag1993: This can be changed to:
`if 'SslPolicy' in listener:`

python/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK.py
```diff
@@ -0,0 +1,436 @@
...
+
+        overall_listeners_eval = 'NON_COMPLIANT'
+
+        for lis in alb_all_listeners:
+            if not is_https_listener(lis):
+                for action in lis['DefaultActions']:
+                    if(action['Type'] == 'redirect' and action['RedirectConfig']['Protocol'] == 'HTTPS'):
+                        overall_listeners_eval = 'COMPLIANT'
+                        continue
+                    else:
```
schirag1993: Since there is a `continue` statement on line 97 (within the "if" block), this else block can be moved to the "for" block

python/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK_test.py
```diff
@@ -0,0 +1,199 @@
...
+        resp_expected.append(build_expected_response('NON_COMPLIANT', 'arn1', annotation="HTTP listener rule must have HTTP to HTTPS redirection action configured"))
+        resp_expected.append(build_expected_response('NON_COMPLIANT', 'arn2', annotation="HTTP listener rule must have HTTP to HTTPS redirection action configured"))
+        assert_successful_evaluation(self, response, resp_expected, 2)
+
+class NotApplicableResourceTest(unittest.TestCase):
+    def test_scenario_1_not_applicable(self):
+        ELBV2_CLIENT_MOCK.describe_load_balancers = MagicMock(
+            return_value={'LoadBalancers': [{'LoadBalancerArn': 'arn1', 'Type': 'other'}]}
+        )
+        ELBV2_CLIENT_MOCK.describe_listeners = MagicMock(
```
schirag1993: `describe_listeners` need not be mocked as the code goes past this API call if there is no ALB

ketanbhut:
@schirag1993 3 of the changes are applied - please review.
Also, suggest where in def evaluate_compliance whitespace needs to be removed. Are you suggesting removing line-breaks? (line #79, #87..)

ketanbhut:
@schirag1993  I have removed line breaks in the evaluate_compliance as discussed offline.

schirag1993:
Code looks good, thank you @ketanbhut. 

@jongogogo, kindly review the PR. Do let us know of any changes to be made.

python/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK.py
```diff
@@ -0,0 +1,427 @@
...
+#############
+# Main Code #
+#############
+
+def evaluate_compliance(event, configuration_item, valid_rule_parameters):
+    evaluations = []
+    alb_client = get_client("elbv2", event)
+    all_elbv2 = get_all_elbv2(alb_client)
+    if not all_elbv2:
+        return [build_evaluation(event['accountId'], 'NOT_APPLICABLE', event, resource_type=ACCOUNT_RESOURCE_TYPE)]
```
jongogogo: No need for a list, and no need for the global variable, since it is not used anywhere else.

python/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK.py
```diff
@@ -0,0 +1,427 @@
...
+    for elb in all_elbv2:
+        if elb['Type'] != 'application':
+            evaluations.append(build_evaluation(elb['LoadBalancerArn'], 'NOT_APPLICABLE', event))
```
jongogogo: No need to add an "NOT_APPLICABLE" evaluation, just `continue`

python/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK.py
```diff
@@ -0,0 +1,427 @@
...
+            continue
+        alb_all_listeners = get_all_listeners(alb_client, elb['LoadBalancerArn'])
+        overall_listeners_eval = 'NON_COMPLIANT'
+        for lis in alb_all_listeners:
+            if not is_https_listener(lis):
+                for action in lis['DefaultActions']:
+                    if(action['Type'] == 'redirect' and action['RedirectConfig']['Protocol'] == 'HTTPS'):
```
jongogogo: parenthesis not required.

python/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK.py
```diff
@@ -0,0 +1,427 @@
...
+                        overall_listeners_eval = 'COMPLIANT'
+                        continue
+                    overall_listeners_eval = 'NON_COMPLIANT'
```
jongogogo: You need a break here, because otherwise if you have 2 http actions, one could overwrite the other.

python/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK.py
```diff
@@ -0,0 +1,427 @@
...
+    if not all_elbv2:
+        return [build_evaluation(event['accountId'], 'NOT_APPLICABLE', event, resource_type=ACCOUNT_RESOURCE_TYPE)]
+    for elb in all_elbv2:
+        if elb['Type'] != 'application':
+            evaluations.append(build_evaluation(elb['LoadBalancerArn'], 'NOT_APPLICABLE', event))
+            continue
+        alb_all_listeners = get_all_listeners(alb_client, elb['LoadBalancerArn'])
+        overall_listeners_eval = 'NON_COMPLIANT'
+        for lis in alb_all_listeners:
+            if not is_https_listener(lis):
```
jongogogo: Do  first a 
```
if is_https_listener(lis):
    overall_listeners_eval = 'COMPLIANT'
    continue
```

It will reduce your indentation afterward.

python/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK.py
ketanbhut: We will have to check NOT_APPLICABLE for 2 scenarios - 1) No ALB/ELB present 2) ALB present but 'Network' (TCP) type - which is also NOT_APPLICABLE as per Gherkin. 
Let me know...

ketanbhut:
Hello @jongogogo  - updated with suggested changes, except one (ALB present but 'Network' (TCP) type scenario being NA) - please review and confirm.

shikharj05:
@jongogogo  @ketanbhut @schirag1993 describe_listeners() API call returns Default
Actions only. As ALB supports multiple rules per listener (path based, host based etc), you need to add describe_rules() to ensure you get action for each rule. 
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elbv2.html#ElasticLoadBalancingv2.Client.describe_rules

This would then match our gherkin #204 
```
Scenario: 3
     Given: There is at least 1 Application Load Balancer
       And: Application Load Balancer has one or more HTTP listeners configured
       And: **At least one HTTP listener rule** does not have HTTP to HTTPS redirection action configured
      Then: Return NON_COMPLIANT
```

Let me know if I missed anything.

jongogogo:
@shikharj05 Good call. @ketanbhut @schirag1993 thoughts?

ketanbhut:
I went through the scenarios where customer would have multiple rules on a given listener, we can definitely check all the rules and I have the code ready for that. It makes sense from the thorough checking perspective as well. However, below observations: 

1. Any Listener will have default action for either HTTP or HTTPS. In case of HTTP listener,  Only default-action will fit into the Gherkin condition of HTTP to HTTPS redirection. 
2. Any additional rule (path/IP/query-string) can have redirection action, but they will not be HTTP to HTTPS redirection action, hence - any listener with any rules other than DefaultAction, will fall into non-compliant. 
3. If we consider any listener with additional rules (path/ip/etc..) would make it NON_COMPLIANT - then we have to consider any HTTPS listener with additional rule *not* redirecting to HTTPS
4. If we do want to enforce that any additional rule also needs to redirect to HTTPS, then we will need to change Gherkin accordingly

Let me know if I am missing anything.

shikharj05:
@ketanbhut I hope our discussion offline solved the above doubts, let's continue with the next commit. Thank you.

ketanbhut:
Thanks @shikharj05. I have updated the rule accordingly. 
@jongogogo, @schirag1993 Please review!

python/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK/ALB_HTTP_TO_HTTPS_REDIRECTION_CHECK.py
```diff
@@ -0,0 +1,434 @@
...
+            if is_https_listener(lis):
+                overall_listeners_eval = 'COMPLIANT'
+                continue
+            rules = get_all_listener_rules(alb_client, lis['ListenerArn'])
+            for rule in rules:
+                for action in rule['Actions']:
+                    if action['Type'] == 'redirect' and action['RedirectConfig']['Protocol'] == 'HTTPS':
+                        overall_listeners_eval = 'COMPLIANT'
+                        continue
+                    break
```
jongogogo: This is not right. The breaks will drive unexpected behavior. 
1. The last break will always be triggered after the first listener.
2. The previous break will trigger after the first rule

ketanbhut:
@jongogogo - Thank you for the inputs. I have corrected the behavior you rightly pointed out and also separated checking rule in a helper function to make it look a little cleaner. Please review and let me know.

PR #238, opened 2019-05-05
STS_ENABLED_IN_REGIONS

virginiachu:

*Description of changes:*
Verify that STS endpoints are enabled only in certain regions.

### DIFF
--- a/python/STS_ENABLED_IN_REGIONS/STS_ENABLE_IN_REGIONS.py
+++ b/python/STS_ENABLED_IN_REGIONS/STS_ENABLE_IN_REGIONS.py
@@ -1,0 +1,425 @@
+# Copyright 2017-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may
+# not use this file except in compliance with the License. A copy of the License is located at
+#
+#        http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for
+# the specific language governing permissions and limitations under the License.
+
+'''
+#####################################
+##           Gherkin               ##
+#####################################
+Rule Name:
+  STS_ENABLE_IN_REGIONS
+Description:
+  Verify that STS endpoints are enabled in whitelisted regions.
+Trigger:
+  Configuration change on AWS::IAM::Role
+Resource Type to report on:
+  AWS::IAM::Role
+Rule Parameters:
+  ActivateRegions
+    Verify Security Token Service is activated in whitelisted region(s).
+Scenarios:
+  Scenario 1:
+  Given: The same number of active and specified regions
+    Then: Return COMPLIANT
+  Scenario 2:
+  Given: There are more active regions then specified.
+    Then: Return NON_COMPLIANT
+  Scenario 3:
+  Given: There are more active regions then specified.
+    Then: Return NON_COMPLIANT
+'''
+import json
+import sys
+import datetime
+import boto3
+import botocore
+
+try:
+    import liblogging
+except ImportError:
+    pass
+
+##############
+# Parameters #
+##############
+
+# Define the default resource to report to Config Rules
+DEFAULT_RESOURCE_TYPE = 'AWS::IAM::Role'
+
+# Set to True to get the lambda to assume the Role attached on the Config Service (useful for cross-account).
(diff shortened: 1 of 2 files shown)


jongogogo:
Thanks!

Can you please :
- add the parameters.json in your PR
- fix pylint errors
- if possible add an exception for the `Session` method in the pylintrc

virginiachu:
Add the words Session, RoleArn, RoleSessionName, and DurationSeconds to pylintrc file as good names to pass the lint errors. These names are needed in order to create mock versions for the boto3 classes such as _boto3.Session()_ .

python/STS_ENABLED_IN_REGIONS/STS_ENABLE_IN_REGIONS.py
```diff
@@ -0,0 +1,425 @@
...
+
+# Other parameters (no change needed)
+CONFIG_ROLE_TIMEOUT_SECONDS = 900
+
+#############
+# Main Code #
+#############
+
+def evaluate_compliance(event, configuration_item, valid_rule_parameters):
+    """Form the evaluation(s) to be return to Config Rules
```
jongogogo: Remove comments

python/STS_ENABLED_IN_REGIONS/STS_ENABLE_IN_REGIONS.py
```diff
@@ -0,0 +1,425 @@
...
+# or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for
+# the specific language governing permissions and limitations under the License.
+
+'''
+#####################################
+##           Gherkin               ##
+#####################################
+Rule Name:
+  STS_ENABLE_IN_REGIONS
```
jongogogo: We must avoid cross region call in AWS Config. It is best you check that a particular region as STS enabled.

python/STS_ENABLED_IN_REGIONS/STS_ENABLE_IN_REGIONS.py
```diff
@@ -0,0 +1,425 @@
...
+
+    Advanced Notes:
+    1 -- if a resource is deleted and generate a configuration change with ResourceDeleted status, the Boilerplate code will put a NOT_APPLICABLE on this resource automatically.
+    2 -- if a None or a list of dictionary is returned, the old evaluation(s) which are not returned in the new evaluation list are returned as NOT_APPLICABLE by the Boilerplate code
+    3 -- if None or an empty string, list or dict is returned, the Boilerplate code will put a "shadow" evaluation to feedback that the evaluation took place properly
+    """
+
+    account_id = valid_rule_parameters["account_id"]
+    role = valid_rule_parameters["role"]
+    session = boto3.Session()
```
jongogogo: Bit worried about this, in the case of a cross account assume role using the ConfigRole role. TThis rule is going to work only in the same account right now, if I understand well.

virginiachu: The current assumed role is not the config’s role but a new role which is the role of the Lambda. The role is working on same account as the config is run. Would need a bit more clarification

python/STS_ENABLED_IN_REGIONS/STS_ENABLE_IN_REGIONS.py
virginiachu: **Questions:**
- When the config rule runs by default it should run in the us-east-1 region.
  Is that correct?
- When the lambda config rule is executed it tests the connectivity to each endpoint.
  Is that considered to be cross region at that point?
- Should the code just be checking if STS is active in the region where the code is running?

python/STS_ENABLED_IN_REGIONS/STS_ENABLE_IN_REGIONS.py
jongogogo: If we move this rule into the console, we need the Config Role to be able to do the job.

python/STS_ENABLED_IN_REGIONS/STS_ENABLE_IN_REGIONS.py
jongogogo: - It is possible to run a config rule, in only 1 region. The rule are independant. The goal is to decouple regions.
- As soon as you do, it makes cross region
```
session = boto3.Session(region_name=region_name)
endpoint_url = f"{sts}{region_name}{domain}"
client = session.client("sts", endpoint_url=endpoint_url)
```
- Yes, if there is no way to get all the data from a single region.

Issue #179, opened 2019-04-03
Configuration change rule for API_GW_NOT_EDGE_OPTIMISED


shikharj05:
Rule on Amazon API Gateway to check that APIs are not edge optimised (API_GW_NOT_EDGE_OPTIMISED).
Please provide comments on the gherkin if any.

---
**Description:**
  Checks that Amazon API Gateway APIs are not edge optimised. The rule returns NON_COMPLIANT if the RestApi's endpointConfiguration is of type EDGE.
**Trigger:**
  Configuration Change on AWS::ApiGateway::RestApi
**Reports on:**
  AWS::ApiGateway::RestApi
**Rule Parameters:**
  None
```
Scenarios:
  Scenario: 1
     Given: The endpointConfiguration type is EDGE in the RestApi configuration item.
     Then: Return NON_COMPLIANT
  Scenario: 2
     Given: The endpointConfiguration type is not EDGE in the RestApi configuration item.
     Then: Return COMPLIANT
```

shikharj05:
Closing in favour of #180

PR #264, opened 2019-06-01
Added Rule for maximum age of Secrets in SM

walkerk1980:

*Issue #, if available:*
Added Rule for maximum age of Secrets in SM

*Description of changes:*
Added Rule for maximum age of Secrets in SM

Trigger Type: Periodic
Accepted Parameters: max_secret_age_days
Example Values: 1, 2, 3,... default 30

requires boto3 >= 1.9.152, botocore >= 1.12.152, will fail without these library versions as pagination of list_secrets() was not supported with the below libraries included in Lambda as of May 21st 2019
Python 3.7 Lambda includes boto3 "1.9.42"
Python 3.6 Lambda includes boto3 "1.7.74"
You must use a custom deployment package that includes the required boto3 and botocore versions
https://aws.amazon.com/premiumsupport/knowledge-center/build-python-lambda-deployment-package/


python/SECRETSMANAGER_MAX_SECRET_AGE.py
```diff
@@ -0,0 +1,77 @@
...
+
+def evaluate_compliance(rule_parameters, secret):
+    if 'max_secret_age_days' in rule_parameters:
+        max_secret_age = datetime.now(timezone.utc) - timedelta(days=int(rule_parameters['max_secret_age_days']))
+    else:
+        max_secret_age = datetime.now(timezone.utc) - timedelta(days=30)
+
+    if 'LastRotatedDate' in secret and datetime.replace(secret['LastRotatedDate'], tzinfo=timezone.utc) > max_secret_age:
+        return "COMPLIANT"
+    if datetime.replace(secret['LastChangedDate'], tzinfo=timezone.utc) > max_secret_age:
```
jongogogo: LastChangedDate is for all changes, including tags, policy, etc. You will need to look at list_secret_version_ids() to see the CreatedDate

python/SECRETSMANAGER_MAX_SECRET_AGE.py
```diff
@@ -0,0 +1,77 @@
...
+    if boto3.__version__ < '1.9.152':
+        print('boto3 version too old for secret pagination')
+        print('boto3 version: ' + boto3.__version__)
+        exit(1)
+
+    now = datetime.now(timezone.utc)
+    if 'ruleParameters' in event:
+        rule_parameters = json.loads(event['ruleParameters'])
+    else:
+        print('No rule paramters found')
```
jongogogo: You have a default value of 30, so it should be optional, not throwing an error.

python/SECRETSMANAGER_MAX_SECRET_AGE.py
```diff
@@ -0,0 +1,77 @@
...
+        result_token = event['resultToken']
+
+    sm_client = boto3.client('secretsmanager')
+    config = boto3.client('config')
+
+    try:
+        paginator = sm_client.get_paginator('list_secrets')
+    except:
+        print('Could not list_secrets')
+        exit(1)
```
jongogogo: This is fine, it just means that the account has no secret. You can call `put_evaluations(Evaluations=[])` to just notify that you checked and it was not applicable. Under the new pricing model, there will be no cost associated to this rule, then.

python/SECRETSMANAGER_MAX_SECRET_AGE.py
walkerk1980: I'm confused about this?? It only throws an error if you were to run it directly in lambda as Config always provides rule parameters in the event even if you don't specify any parameters. This should only throw an error when not run by ConfigService.

python/SECRETSMANAGER_MAX_SECRET_AGE.py
walkerk1980: noted and changed

walkerk1980: Actually, I have the paginator in a try block for error handling of old libs only, if libs too old get_paginator('list_secrets') throws 'This operation does not support pagination'.

If there are no Secrets in Account an empty array is returned, no exception is thrown.

What are you looking for me to change here? I can just remove the try block altogether.

jongogogo: My goal: the behavior for having no secret should be to put_evaluation on an empty list. I am not really worried about the paginator, since you are already checking boto3 version.

python/SECRETSMANAGER_MAX_SECRET_AGE.py
jongogogo: The idea is that if the customer does not provide any input parameters, then the rule should default to 30. The rule (https://docs.aws.amazon.com/config/latest/developerguide/acm-certificate-expiration-check.html) works this way, if you put no parameter, it defaults to 14 days.

python/SECRETSMANAGER_MAX_SECRET_AGE.py
```diff
@@ -0,0 +1,77 @@
+#
```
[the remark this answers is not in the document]
walkerk1980: The rules code by the RDK relies on getting configuration items from the invoking event correct?
Not sure how I can use that for resource types that are not supported by Config Service.

def lambda_handler(event, context):
...
            configuration_item = get_configuration_item(invoking_event)
            if is_applicable(configuration_item, event):
                compliance_result = evaluate_compliance(event, configuration_item, valid_rule_parameters)
...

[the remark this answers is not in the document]
jongogogo: The RDK works for periodic rule as well. If it is a periodic rule, the `configuration_item` will be an empty dict.

python/SECRETSMANAGER_MAX_SECRET_AGE.py
walkerk1980: I understand about defaulting to 30 for a value within rule_parameters, but that has nothing to do with this portion of code. This portion of code will never raise an exception when run by a Config Rule.

jongogogo: L49 as an `exit(1)`

python/SECRETSMANAGER_MAX_SECRET_AGE.py
```diff
@@ -0,0 +1,77 @@
...
+# the required boto3 and botocore versions
+# https://aws.amazon.com/premiumsupport/knowledge-center/build-python-lambda-deployment-package/
+
+import json
+from datetime import datetime, timedelta, timezone
+import boto3
+
+def evaluate_compliance(rule_parameters, secret):
+    if 'max_secret_age_days' in rule_parameters:
+        max_secret_age = datetime.now(timezone.utc) - timedelta(days=int(rule_parameters['max_secret_age_days']))
```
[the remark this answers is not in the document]
walkerk1980: added suggested validation

python/SECRETSMANAGER_MAX_SECRET_AGE.py
walkerk1980: used versions instead

python/SECRETSMANAGER_MAX_SECRET_AGE.py
walkerk1980: This is what I'm saying though, L49 makes rule_parameters mandatory, not rule_parameters['max_secret_age_days'], which is optional.

walkerk1980:
> ************* Module SECRETSMANAGER_MAX_SECRET_AGE_test
> SECRETSMANAGER_MAX_SECRET_AGE/SECRETSMANAGER_MAX_SECRET_AGE_test.py:27:8: R1705: Unnecessary "elif" after "return" (no-else-return)
> SECRETSMANAGER_MAX_SECRET_AGE/SECRETSMANAGER_MAX_SECRET_AGE_test.py:26:4: R0201: Method could be a function (no-self-use)
> SECRETSMANAGER_MAX_SECRET_AGE/SECRETSMANAGER_MAX_SECRET_AGE_test.py:48:8: W1503: Redundant use of assertTrue with constant value True (redundant-unittest-assert)
> SECRETSMANAGER_MAX_SECRET_AGE/SECRETSMANAGER_MAX_SECRET_AGE_test.py:6:4: W0611: Unused import mock (unused-import)
> SECRETSMANAGER_MAX_SECRET_AGE/SECRETSMANAGER_MAX_SECRET_AGE_test.py:9:0: W0611: Unused ClientError imported from botocore.exceptions (unused-import)
> ************* Module SECRETSMANAGER_MAX_SECRET_AGE
> SECRETSMANAGER_MAX_SECRET_AGE/SECRETSMANAGER_MAX_SECRET_AGE.py:231:4: R1705: Unnecessary "elif" after "return" (no-else-return)
> SECRETSMANAGER_MAX_SECRET_AGE/SECRETSMANAGER_MAX_SECRET_AGE.py:249:12: R1714: Consider merging these comparisons with "in" to "status in ('OK', 'ResourceDiscovered')" (consider-using-in)
> 
> Your code has been rated at 9.99/10

The RDK generated code causes pylint-github-aws-config-rules to fail?

python/SECRETSMANAGER_MAX_SECRET_AGE.py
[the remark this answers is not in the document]
walkerk1980: Not sure I used it as intended, but Config rule based on RDK does work, I had to './venv/bin/pip3 install -r requirements.txt -t .' in the rule directory to get pagination to work of course.

jongogogo:
Unfortunately yes, we need to update the template. I thought we did though. You are using the python3.6 template? @michaelborchert

walkerk1980:
> Unfortunately yes, we need to update the template. I thought we did though. You are using the python3.6 template? @michaelborchert

I'm using 3.7, is the 3.6 template more up-to-date?
I can switch to that and merge my changes to it instead.

jongogogo:
Yes! That's must be it. Can you switch to the 3.6, please?

walkerk1980:
Diffed:
https://github.com/awslabs/aws-config-rdk/blob/master/rdk/template/runtime/python3.7/rule_code.py
https://github.com/awslabs/aws-config-rdk/blob/master/rdk/template/runtime/python3.6/rule_code.py
3.7 is not in sync with 3.6, integrated changes from 3.6 and it's tests which passes pylint-github-aws-config-rules

python/SECRETSMANAGER_MAX_SECRET_AGE.py
walkerk1980: removed paginator from try block entirely

python/SECRETSMANAGER_MAX_SECRET_AGE.py
walkerk1980: I removed this entirely and as the RDK code already has code for this.

walkerk1980:
That should take care of that.

https://github.com/awslabs/aws-config-rdk/pull/179 Sync out of date python3.7 rule_code.py and rule_test.py templates w/3.6 #179

python/SECRETSMANAGER_MAX_SECRET_AGE/SECRETSMANAGER_MAX_SECRET_AGE.py
```diff
@@ -0,0 +1,424 @@
...
+    #now = datetime.now(timezone.utc)
+    delta = timedelta(days=int(valid_rule_parameters.get('max_secret_age_days') if valid_rule_parameters.get('max_secret_age_days') else '30'))
+    max_secret_age = NOW - delta
+
+    if secret.get('LastRotatedDate'):
+        if datetime.replace(secret.get('LastRotatedDate'), tzinfo=timezone.utc) > max_secret_age:
+            return 'COMPLIANT'
+        return 'NON_COMPLIANT'
+
+    secret_versions = AWS_SECRETSMANAGER_CLIENT.list_secret_version_ids(
```
jongogogo: Can you please add a comment that we don't need a paginator, because this case should only be possible when a secret has never been rotated, and therefore it must have only 1 version.

python/SECRETSMANAGER_MAX_SECRET_AGE/SECRETSMANAGER_MAX_SECRET_AGE.py
```diff
@@ -0,0 +1,424 @@
...
+CONFIG_ROLE_TIMEOUT_SECONDS = 900
+
+#############
+# Main Code #
+#############
+
+
+def evaluate_secret_compliance(valid_rule_parameters, secret):
+    #now = datetime.now(timezone.utc)
+    delta = timedelta(days=int(valid_rule_parameters.get('max_secret_age_days') if valid_rule_parameters.get('max_secret_age_days') else '30'))
```
jongogogo: Do the logic of replacement in the evaluate_parameter()

python/SECRETSMANAGER_MAX_SECRET_AGE/SECRETSMANAGER_MAX_SECRET_AGE.py
```diff
@@ -0,0 +1,424 @@
...
+    max_secret_age = NOW - delta
```
jongogogo: No need of a global variable.

python/SECRETSMANAGER_MAX_SECRET_AGE/SECRETSMANAGER_MAX_SECRET_AGE.py
```diff
@@ -0,0 +1,424 @@
...
+    Return:
+    anything suitable for the evaluate_compliance()
+
+    Keyword arguments:
+    rule_parameters -- the Key/Value dictionary of the Config Rules parameters
+    """
+    if 'max_secret_age_days' in rule_parameters and int(rule_parameters.get('max_secret_age_days')) > int('999999999'):
+        raise ValueError('max_secret_age_days must be less than 999999999')
+    valid_rule_parameters = rule_parameters
+    return valid_rule_parameters
```
jongogogo: Add a global variable on the top `DEFAULT_MAX_SECRET_AGE_DAYS = 30`

then something like this
```
try:
     max_secret_age_days = int(rule_parameters.get('max_secret_age_days', DEFAULT_MAX_SECRET_AGE_DAYS)
except TypeError:
    raise ValueError("...must be an integer")

if max_secret_age_days > 1E9-1:
   raise ValueError("...must be less than ...")

return {"max_secret_age_days" : max_secret_age_days }

python/SECRETSMANAGER_MAX_SECRET_AGE/SECRETSMANAGER_MAX_SECRET_AGE.py
walkerk1980: `    # Pagination of this API call is not needed as this API is only called if Secret has never been rotated
    # This should always return only a single VersionId with VersionLabel AWSCURRENT`

python/SECRETSMANAGER_MAX_SECRET_AGE/SECRETSMANAGER_MAX_SECRET_AGE.py
walkerk1980: Added global variable and moved logic to evaluate_parameters()

python/SECRETSMANAGER_MAX_SECRET_AGE/SECRETSMANAGER_MAX_SECRET_AGE.py
walkerk1980: Added global variable and moved logic to evaluate_parameters()

python/SECRETSMANAGER_MAX_SECRET_AGE/SECRETSMANAGER_MAX_SECRET_AGE.py
walkerk1980: I removed the Global variable, my logic here though is that without calling datetime.now() outside of evaluate_secret_compliance() each secret is evaluated by a different time. It could be possible for two secrets of around the same age to be evaluated differently if the rule is run within a couple minutes of the 24 hour point if the rule takes long enough to run. This is not very important in reality.

PR #272, opened 2019-06-14
[EMR_SECURITY_GROUPS_RESTRICTED] replaced ec2 api call with config api

kritijha:


kritijha:
@jongogogo review when you can :)

python/EMR_SECURITY_GROUPS_RESTRICTED/EMR_SECURITY_GROUPS_RESTRICTED.py
```diff
@@ -116,29 +116,39 @@ def get_sg_cluster_mapping(cluster_list, emr_client):
...
-            elif is_sg_open_ipv6(security_group):
-                open_sg_list.add(security_group['GroupId'])
+    resource_keys = []
+    for security_group in sg_list:
+        resource_key = {
+            "resourceType": "AWS::EC2::SecurityGroup",
+            "resourceId": security_group
+        }
+        resource_keys.append(resource_key)
+    res = config_client.batch_get_resource_config(resourceKeys=resource_keys)
```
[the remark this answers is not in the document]
kritijha: Made changes as per comments 1 & 3
Regarding 2, consider the following:
i) If the resource has just been created, it won't be picked up by Config for some time. During this time, no configuration item will be returned for the resource.
ii) If the resource has been deleted recently, it will be a part of Config's resources for some time before it is marked as deleted. Post then, no config item will be returned.

In both the cases, no error will be thrown. Consequently, I feel we can do away with error handling and leave the change to be reflected in the next execution of the Config rule.

@jongogogo What are your views?

kritijha:
Not sure why pylint exceptions are being thrown for other code files 🤷🏻‍♀️

python/EMR_SECURITY_GROUPS_RESTRICTED/EMR_SECURITY_GROUPS_RESTRICTED.py
```diff
@@ -116,36 +115,52 @@ def get_sg_cluster_mapping(cluster_list, emr_client):
...
+        resource_key = {
+            "resourceType": "AWS::EC2::SecurityGroup",
+            "resourceId": security_group
+        }
+        resource_keys.append(resource_key)
+    while resource_keys:
+        configuration_items = get_config_items(config_client, resource_keys[:100])
+        del resource_keys[:100]
+    for item in configuration_items:
+        if is_sg_open_ipv4(json.loads(item["configuration"])):
```
jongogogo: change the function to better handle the json load and make more readable.

```
config_item = json.loads(item["configuration"])
if is_sg_open(config_item):
     open_sg_list.add(config_item['groupId'])

def is_sg_open(config_item):
    return is_sg_open_ipv4(config_item) or is_sg_open_ipv6(config_item)
```

jongogogo:
Don't worry about the pylint error. I changed the pylintrc to include a new version of the VPC_SG_... rule. Hence the problem :)

python/EMR_SECURITY_GROUPS_RESTRICTED/EMR_SECURITY_GROUPS_RESTRICTED.py
```diff
@@ -116,36 +115,52 @@ def get_sg_cluster_mapping(cluster_list, emr_client):
...
     return False
 
 def evaluate_parameters(rule_parameters):
     valid_rule_parameters = rule_parameters
     return valid_rule_parameters
 
+def get_config_items(config_client, resource_keys):
+    res = config_client.batch_get_resource_config(resourceKeys=resource_keys)
+    configuration_items = res["baseConfigurationItems"]
+    while res["unprocessedResourceKeys"]:
```
jongogogo: You are risking an infinite loop here: if the cluster get created, then the rule run before the sg is registered in Config. The sg will remain into the said 'unprocessedResourceKeys' for up to 15 minutes (or more).
I suggest: after 3 tries -> mark the sg as "unregistered", and do not post any information about the EMR cluster(s) containing it.

kritijha: That's just the thing! If the SG doesn't exist, it doesn't turn up in the unprocessed items list. It just seems to be skipped. The same happens if you provide some random bogus resource id. 
For example:
```
aws configservice batch-get-resource-config --resource-keys resourceType=AWS::EC2::SecurityGroup,resourceId=sg-0ff787b5ade90a1243
{
    "baseConfigurationItems": [],
    "unprocessedResourceKeys": []
}
```

kritijha:
@jongogogo better now? I've offloaded all the config item related work to the method get_config_items(). If needed, we should be able to reuse it across other rules.

python/EMR_SECURITY_GROUPS_RESTRICTED/EMR_SECURITY_GROUPS_RESTRICTED.py
jongogogo: Oooh, I see now your point. Well there is the easy way, and we run with what is today. Or we need to compare resource_keys with the configuration_item. I think let's start with what we have first. We can move into CR.

PR #290, opened 2019-09-25
Add deployment rdk for new managed rules

jongogogo:

*Issue #, if available:*

*Description of changes:*


RobReus:
Why would you delete the source code of the managed rules? The source code was very useful in the scenario where a small change was need to fit a specific use-case. Where can we view the source code of these managed rules now?

jongogogo:
Good point, happy to let it there. I'll do a revert + merge soon.

RobReus:
@jongogogo this has not been reverted yet. Could you be so kind to do the revert + merge? Thanks!

PR #307, opened 2020-01-04
Custom Rule for API Gateway to require an Authorizer

IanMeyers:

*Issue #, if available:*

*Description of changes:*

Customer often want to have completely open public API's, but in other cases would like to apply a rule that requires that their API's be secured in a common way, such as through Web Identity Federation, or by using AWS IAM. This custom rule runs against API Gateway Stages and checks that they have Authorizers attached. Furthermore, customers can specify what types of authorizer they will allow on their API's, ensuring that a consistent configuration is applied to API's serving back end data that must be secured.


IanMeyers:
Not sure about these CodeBuild checks. The line numbers don't line up with advice correctly, and it's failing checks on this PR from code that isn't included.

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
```diff
@@ -0,0 +1,505 @@
...
+        valid_authorizers = valid_rule_parameters.get("RequireAuthTypes")
+
+    if "configurationItem" in config_rule_event:
+        configuration_item = config_rule_event.get("configurationItem")
+        api_id = configuration_item.get("resourceId").split("/")[2]
+        api_region = configuration_item.get("awsRegion")
+
+        global _client_cache
+        if _client_cache is None:
+            _client_cache = {api_region: get_client(service="apigateway", event=event, region=api_region)}
```
jongogogo: No need to bother with the region, just do
```
client_apigw = get_client('apigateway', event)
```

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
```diff
@@ -0,0 +1,505 @@
...
+
+        api_client = _client_cache.get(api_region)
+
+        # get all resources for this API
+        resources = api_client.get_resources(
```
jongogogo: You must either rely solely on configuration_item or create a periodic rule. Otherwise the rule will not trigger on changes.

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
```diff
@@ -0,0 +1,505 @@
...
+        api_id = configuration_item.get("resourceId").split("/")[2]
+        api_region = configuration_item.get("awsRegion")
+
+        global _client_cache
+        if _client_cache is None:
+            _client_cache = {api_region: get_client(service="apigateway", event=event, region=api_region)}
+
+        api_client = _client_cache.get(api_region)
+
+        # get all resources for this API
```
jongogogo: You are getting "only" 500 of them

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
```diff
@@ -0,0 +1,505 @@
...
+
+            for resource in resources.get("items"):
+                if "resourceMethods" in resource:
+                    for method, method_config in resource.get("resourceMethods").items():
+                        resource_method = f"{api_id}:{method}:{resource.get('path')}"
+
+                        auth = method_config.get("authorizationType", None)
+
+                        if auth is None or auth == 'NONE':
+                            methods_compliant = False
```
jongogogo: As soon as you have a False, you should break the loop to limit API usage

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
```diff
@@ -0,0 +1,505 @@
...
+                            no_auth += 1
+                            print(f"No Authorizer found on {resource_method}")
+                        else:
+                            if valid_authorizers is not None:
```
jongogogo: `if valid_authorizers:` is more pythonic

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
```diff
@@ -0,0 +1,505 @@
...
+
+            # add the summary evaluation for the stage
+            if methods_compliant:
+                _add_response(stage=configuration_item.get("resourceId"), status="COMPLIANT",
+                              event=event,
+                              comment="All Stage Resource Methods are compliant with Authorizer Policy")
+            else:
+                _add_response(stage=configuration_item.get("resourceId"), status="NON_COMPLIANT",
+                              event=event,
+                              comment=f"Found {no_auth} methods without Authorizers, {invalid_auth} methods without Valid Authorizers, and {ok_auth} with correct Authorizers")
```
jongogogo: Those value introduce complexity and API usage which is not required. You can state, "at least 1 method is not using a valid authorizers"

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
```diff
@@ -0,0 +1,505 @@
...
+#############
+# Main Code #
+#############
+
+_client_cache = None
+_config_rule_response = []
+API_GATEWAY_STAGE_TYPE = "AWS::ApiGateway::Stage"
+
+
+def _add_response(stage, status, event, comment):
```
jongogogo: This seems superfluous, I suggest to use build_evaluation() and append only

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
```diff
@@ -0,0 +1,505 @@
...
+# Other parameters (no change needed)
+CONFIG_ROLE_TIMEOUT_SECONDS = 900
+
+#############
+# Main Code #
+#############
+
+_client_cache = None
+_config_rule_response = []
+API_GATEWAY_STAGE_TYPE = "AWS::ApiGateway::Stage"
```
jongogogo: Build_evaluation use DEFAULT_RESOURCE_TYPE variable as a default for the resource type. You can just change it and remove this variable.

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
```diff
@@ -0,0 +1,505 @@
...
+    event -- the event variable given in the lambda handler
+    configuration_item -- the configurationItem dictionary in the invokingEvent
+    valid_rule_parameters -- the output of the evaluate_parameters() representing validated parameters of the Config Rule
+
+    Advanced Notes:
+    1 -- if a resource is deleted and generate a configuration change with ResourceDeleted status, the Boilerplate code will put a NOT_APPLICABLE on this resource automatically.
+    2 -- if a None or a list of dictionary is returned, the old evaluation(s) which are not returned in the new evaluation list are returned as NOT_APPLICABLE by the Boilerplate code
+    3 -- if None or an empty string, list or dict is returned, the Boilerplate code will put a "shadow" evaluation to feedback that the evaluation took place properly
+    """
+    config_rule_event = json.loads(event.get("invokingEvent"))
```
jongogogo: You can just use the configuration_item which is already passed, not need to get it again.

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
IanMeyers: Agreed

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
IanMeyers: Why is that? Not providing a region will default this to us-east-1 no? That's not going to work everywhere...

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
IanMeyers: Yep - adding pagination semantics to load everything.

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
IanMeyers: ack

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
IanMeyers: Understood. Would be nice if RDK docs are more clear about what this value contains.

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
IanMeyers: Was using it for the resource type, but will use default resource type now so agree.

python/API_GW_REQUIRE_AUTHORIZER/API_GW_REQUIRE_AUTHORIZER.py
IanMeyers: In this case, the Stage is updated, and then I load the resources for the stage, which are not part of the configuration item. This would be the desired behaviour.

PR #318, opened 2020-02-03
updated logic and tests for sagemaker

sowmitranalla:

*Description of changes:*

The API response from the SageMaker client for `describe_notebook_instance` is below. The value at `KmsKeyId` is just the Key Id, not the ARN as is expected in the config rule code. Due to this difference, compliant notebook instances are being flagged as non compliant. 

The updated logic checks to make sure the `KmsKeyId` string exists in at least one of the supplied ARNs and if it does, the instance is compliant. The test mocks were also updated to reflect the API response.

```{
    "NotebookInstanceArn": "arn:aws:sagemaker:us-east-1:xxx:notebook-instance/compliant-sagemaker",
    "NotebookInstanceName": "compliant-sagemaker",
    "NotebookInstanceStatus": "InService",
    "Url": "compliant-sagemaker.notebook.us-east-1.sagemaker.aws",
    "InstanceType": "ml.t2.medium",
    "SubnetId": "subnet-xxx",
    "SecurityGroups": [
        "sg-xxxx"
    ],
    "RoleArn": "arn:aws:iam::xxx:role/sagemaker-role",
    "KmsKeyId": "3xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxe",
    "NetworkInterfaceId": "eni-xxx",
    "LastModifiedTime": 1580509333.953,
    "CreationTime": 1580509123.302,
    "DirectInternetAccess": "Disabled",
    "VolumeSizeInGB": 5,
    "RootAccess": "Enabled"
}```

### DIFF
--- a/python/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED.py
+++ b/python/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED.py
@@ -65,6 +65,7 @@ import datetime
 import re
 import boto3
 import botocore
+from functools import reduce
 
 try:
     import liblogging
@@ -107,7 +108,7 @@ def evaluate_compliance(event, configuration_item, valid_rule_parameters):
         elif not valid_rule_parameters:
             evaluations.append(build_evaluation(notebook_instance['NotebookInstanceName'], 'COMPLIANT', event))
         #SCENARIO 5: KMS key specified for Amazon SageMaker Notebook Instance matches keyArn in rule parameter.
-        elif notebook_instance_description['KmsKeyId'] in valid_rule_parameters:
+        elif reduce((lambda x, y: x or y), [notebook_instance_description['KmsKeyId'] in keyId for keyId in valid_rule_parameters]):
             evaluations.append(build_evaluation(notebook_instance['NotebookInstanceName'], 'COMPLIANT', event))
         #SCENARIO 4: KMS key specified for Amazon SageMaker Notebook Instance does not match keyArn in rule parameter.
         else:
--- a/python/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED.py
+++ b/python/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED.py
@@ -63,9 +63,9 @@ import json
 import sys
 import datetime
 import re
+from functools import reduce
 import boto3
 import botocore
-from functools import reduce
 
 try:
     import liblogging
--- a/python/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED_test.py
+++ b/python/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED_test.py
@@ -53,9 +53,9 @@ class ComplianceTest(unittest.TestCase):
     no_notebook_instances_list = {"NotebookInstances": []}
     notebook_instances_list_mixed = [{'NotebookInstances': [{'NotebookInstanceName': 'trial12'}, {'NotebookInstanceName': 'trial123'}, {'NotebookInstanceName': 'trial1234'}]}]
 
-    described_notebook_instances = [{'NotebookInstanceName': 'trial12', 'KmsKeyId': 'arn:aws:kms:us-east-1:123456789012:key/7af97db7-f6a3-4d0a-87b9-a2737b54856d'}, {'NotebookInstanceName': 'trial123', 'KmsKeyId': 'arn:aws:kms:us-east-1:123456789012:key/7af97db7-f6a3-4d0a-87b9-a2737b54856d'}]
+    described_notebook_instances = [{'NotebookInstanceName': 'trial12', 'KmsKeyId': '7af97db7-f6a3-4d0a-87b9-a2737b54856d'}, {'NotebookInstanceName': 'trial123', 'KmsKeyId': '7af97db7-f6a3-4d0a-87b9-a2737b54856d'}]
     described_notebooks_no_key = [{'NotebookInstanceName': 'trial12'}, {'NotebookInstanceName': 'trial123'}]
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: python/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED.py [hunk 1]
```diff
@@ -107,7 +108,7 @@ def evaluate_compliance(event, configuration_item, valid_rule_parameters):
         elif not valid_rule_parameters:
             evaluations.append(build_evaluation(notebook_instance['NotebookInstanceName'], 'COMPLIANT', event))
         #SCENARIO 5: KMS key specified for Amazon SageMaker Notebook Instance matches keyArn in rule parameter.
-        elif notebook_instance_description['KmsKeyId'] in valid_rule_parameters:
+        elif reduce((lambda x, y: x or y), [notebook_instance_description['KmsKeyId'] in keyId for keyId in valid_rule_parameters]):
```
jongogogo: Please explain the value of doing this. "valid_rule_parameters" is a list of key and the notebook_instance_description['KmsKeyId'] is a single key.

python/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED_test.py
```diff
@@ -53,9 +53,9 @@ class ComplianceTest(unittest.TestCase):
     no_notebook_instances_list = {"NotebookInstances": []}
     notebook_instances_list_mixed = [{'NotebookInstances': [{'NotebookInstanceName': 'trial12'}, {'NotebookInstanceName': 'trial123'}, {'NotebookInstanceName': 'trial1234'}]}]
 
-    described_notebook_instances = [{'NotebookInstanceName': 'trial12', 'KmsKeyId': 'arn:aws:kms:us-east-1:123456789012:key/7af97db7-f6a3-4d0a-87b9-a2737b54856d'}, {'NotebookInstanceName': 'trial123', 'KmsKeyId': 'arn:aws:kms:us-east-1:123456789012:key/7af97db7-f6a3-4d0a-87b9-a2737b54856d'}]
+    described_notebook_instances = [{'NotebookInstanceName': 'trial12', 'KmsKeyId': '7af97db7-f6a3-4d0a-87b9-a2737b54856d'}, {'NotebookInstanceName': 'trial123', 'KmsKeyId': '7af97db7-f6a3-4d0a-87b9-a2737b54856d'}]
```
jongogogo: Good catch!

jongogogo: I just run the test on an sagemaker notebook. It is actually the KMS key ARN which is given by sagemaker_client.describe_notebook_instance(), not the KMS key ID. There is a discrepancy between the actual value, the key and the documentation. But bottom line, the rule code should work.
Where did you run your test? I am currently running my test in ap-southeast-1.

sowmitranalla: Hmm..that's very strange. I ran my test in us-east-1 and got KMS key ID and haven't tried it in other regions.

python/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED.py
sowmitranalla: I am assuming that the KMS key ID is given at the `notebook_instance_description['KmsKeyId']` and this is a substring of the ARN that's passed into the rule parameters. The reduce function will be given an array of boolean values (ex, `[False, False, True]`) and it'll return True if _one_ item in the array is true. This way, we can see if an KMS key ID exists anywhere in the list of KMS ARNs.

jongogogo: valid_rule_parameters is a list of all the KmsKeyARN that are authorized. If the instance key matches any of those key, it is considered compliant. The reduce is not necessary here.

python/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED/SAGEMAKER_NOTEBOOK_KMS_CONFIGURED_test.py
jongogogo: I just run my test in us-east-1. I get a KmsKeyArn as well. Can you please tell the step to reproduce your test?

I am running a python3.7 lambda function

```
import json
import boto3

def lambda_handler(event, context):
    sagemaker_client = boto3.client('sagemaker')
    notebook_instance_description = sagemaker_client.describe_notebook_instance(NotebookInstanceName="test-for-rule")
    return {
        'statusCode': 200,
        'body': str(notebook_instance_description)
    }
```

sowmitranalla: This PR doesn't seem like it's necessary.

I had created the SageMaker instance through terraform and used the KMS key ID when creating the instance the [CreateNotebookInstance API](https://docs.aws.amazon.com/sagemaker/latest/dg/API_CreateNotebookInstance.html#SageMaker-CreateNotebookInstance-request-KmsKeyId) accepted just the ID. Because I gave the key ID, that's what I saw I the response. I just created the instance again with the KMS ARN and the rule worked as expected.

Never went down this rabbit hole but do you know where I can see the repo that handles the REST APIs for SageMaker or is that entirely internal to AWS?

jongogogo: Ok cool.

As far as I know, we didn't opensource the APIs handlers as such. After all, "APIs are forever" :)
You can see the read-only standard definition file in our SDKs though. For example, boto: https://github.com/boto/botocore/blob/develop/botocore/data/sagemaker/2017-07-24/service-2.json

sowmitranalla: Good to know. Thanks for all your quick responses!

PR #320, opened 2020-02-06
Added rule to check API Gateway for custom domain name

jchua06:

*Issue #, if available:*

*Description of changes:*
Added a new AWS config rule to check whether API Gateway has a custom domain name.


python/API_GW_CUSTOMDOMAIN_CHECK/API_GW_CUSTOMDOMAIN_CHECK.py
```diff
@@ -0,0 +1,395 @@
...
+# Main Code #
+#############
+
+def evaluate_compliance(event, configuration_item, valid_rule_parameters):
+    apigw = get_client('apigateway', event)
+    customdomains = []
+    evaluations = []
+    count = 0
+    # make the api call
+    response = apigw.get_domain_names()
```
jongogogo: You need a loop to agreggate all the domain_names. With this call you will get only 25.

python/API_GW_CUSTOMDOMAIN_CHECK/API_GW_CUSTOMDOMAIN_CHECK.py
```diff
@@ -0,0 +1,395 @@
...
+def evaluate_parameters(rule_parameters):
+    """Evaluate the rule parameters dictionary validity. Raise a ValueError for invalid parameters.
+
+    Return:
+    anything suitable for the evaluate_compliance()
+
+    Keyword arguments:
+    rule_parameters -- the Key/Value dictionary of the Config Rules parameters
+    """
+    valid_rule_parameters = rule_parameters
```
jongogogo: You might consider to have a parameter to check if the custom domain is correct. Something like a list of regex patterns

python/API_GW_CUSTOMDOMAIN_CHECK/API_GW_CUSTOMDOMAIN_CHECK_test.py
```diff
@@ -0,0 +1,202 @@
...
+        "items":[
+            {
+                "domainNameStatus": "AVAILABLE",
+                "domainName": "api.sitepep.com",
+                "endpointConfiguration": {
+                    "types": [
+                        "REGIONAL"
+                    ]
+                },
+                "regionalCertificateArn": "arn:aws:acm:us-east-2:052767133087:certificate/bdd25578-326f-4783-913a-2acef52be012",
```
jongogogo: Put a fake Account ID 12345678912 or 112233445566

PR #329, opened 2020-05-18
Added custom rule to remove public CIDR block from Security Group

soroushatarod:

*Issue #, if available:*

*Description of changes:*
There are environments such as the production which we do not want to have a security group with public CIDR block except certain Security Groups only. This rule will remove the public accessible CIDR block from the security group.


soroushatarod:
@jongogogo could you please review? Not quite sure how to get the module name correct. I have added it to the pylintrc ignore

fuellbie:
Hi, according to the regex, the Module name must consist of capital letters. Try EC2_SECURITY_GROUP_REMOVE_PUBLIC_CIDR!

PR #352, opened 2020-12-15
add rule for checking ecr repo image scan on push setting

suhussai:

*Issue #, if available:*

*Description of changes:*
Add python3.7 rule for checking image scan setting on ecr repos.

### DIFF
--- a/python/ECR_REPOSITORY_SCAN_ON_PUSH_CHECK/ECR_REPOSITORY_SCAN_ON_PUSH_CHECK.py
+++ b/python/ECR_REPOSITORY_SCAN_ON_PUSH_CHECK/ECR_REPOSITORY_SCAN_ON_PUSH_CHECK.py
@@ -125,12 +125,18 @@ def evaluate_parameters(rule_parameters):
     Keyword arguments:
     rule_parameters -- the Key/Value dictionary of the Config Rules parameters
     """
+    if parameters_exist(rule_parameters):
+        raise ValueError('This rule is not configured to take any input parameters.')
+
     valid_rule_parameters = rule_parameters
     return valid_rule_parameters
 
 ####################
 # Helper Functions #
 ####################
+
+def parameters_exist(parameters):
+    return len(parameters) > 0
 
 # Build an error to be displayed in the logs when the parameter is invalid.
 def build_parameters_value_error_response(ex):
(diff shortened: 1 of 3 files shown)


python/ECR_REPOSITORY_SCAN_ON_PUSH_CHECK/ECR_REPOSITORY_SCAN_ON_PUSH_CHECK.py
```diff
@@ -0,0 +1,452 @@
...
+def evaluate_parameters(rule_parameters):
+    """Evaluate the rule parameters dictionary validity. Raise a ValueError for invalid parameters.
+
+    Return:
+    anything suitable for the evaluate_compliance()
+
+    Keyword arguments:
+    rule_parameters -- the Key/Value dictionary of the Config Rules parameters
+    """
+    valid_rule_parameters = rule_parameters
```
jschwarzwalder, line 128: Are you doing any validation checks here?

suhussai: I wasn't sure what to check for. The rule isn't designed to take in a parameter.

I was just looking at another rule that does something similar (i.e. obtain a list of resources of a single type and iterate over each one to make sure that a given option is enabled) and found this one: https://github.com/awslabs/aws-config-rules/blob/master/python/EKS_LOGGING_CHECK/EKS_LOGGING_CHECK.py and it looks like they are doing something similar:
```python
def evaluate_parameters(rule_parameters):
    """Evaluate the rule parameters dictionary validity. Raise a ValueError for invalid parameters.
    Return:
    anything suitable for the evaluate_compliance()
    Keyword arguments:
    rule_parameters -- the Key/Value dictionary of the Config Rules parameters
    """
    valid_rule_parameters = rule_parameters
    return valid_rule_parameters
```

Let me know if you have any suggestions. Should I be validating that no parameters have been passed in?

jschwarzwalder: it would be good to add another evaluation that raises a ValueError if there are any parameters. 

```
def parameters_exist(parameters):
    return len(parameters) != 0
```

suhussai: I added a check for parameters in the rule and included a test to assert an error when that happens.

PR #420, opened 2024-03-04
Add config rule for checking on-demand EC2 license running on dedicat…

jonamadi:
…ed host

*Issue #, if available:*

*Description of changes:*

Adds Python AWS Config rule which checks if an EC2 instance running on Dedicated Hosts is not running as BYOL. This helps customers detect potential misconfiguration issues where they may be double-paying OS licenses (e.g. they are running a Windows License-Included AMI on Dedicated Hosts and mistakenly assuming their Enterprise license is covering it, or have a valid BYOL license but did not convert the license using AWS License Manager when moving to a dedicated host).


derekkzh (approved):
peolivei@ approved, but does not have github access.

PR #433, opened 2025-01-06
Removing S3_BUCKET_LEVEL_PUBLIC_ACCESS_PROHIBITED which is redundant to limit rules to 130

pallavi03:

*Issue #, if available:*
Last [commit](https://code.amazon.com/packages/AwsConformsTemplates/commits/52a23701dc8f8545f21ecfcf4ac0af5784fef631) increase number of rules in some templates to 131 which is above limit of 130 rules in a CPack. 

*Description of changes:*
After checking these templates found that [S3_BUCKET_LEVEL_PUBLIC_ACCESS_PROHIBITED](https://docs.aws.amazon.com/config/latest/developerguide/s3-bucket-level-public-access-prohibited.html) can be removed as the check is covered by [S3_BUCKET_PUBLIC_READ_PROHIBITED](https://docs.aws.amazon.com/config/latest/developerguide/s3-bucket-public-read-prohibited.html) and [S3_BUCKET_PUBLIC_WRITE_PROHIBITED](https://docs.aws.amazon.com/config/latest/developerguide/s3-bucket-public-write-prohibited.html)

### DIFF
--- a/aws-config-conformance-packs/Operational-Best-Practices-For-NIST-800-181.yaml
+++ b/aws-config-conformance-packs/Operational-Best-Practices-For-NIST-800-181.yaml
@@ -1216,16 +1216,6 @@ Resources:
         Owner: AWS
         SourceIdentifier: S3_ACCOUNT_LEVEL_PUBLIC_ACCESS_BLOCKS_PERIODIC
     Type: AWS::Config::ConfigRule
-  S3BucketLevelPublicAccessProhibited:
-    Properties:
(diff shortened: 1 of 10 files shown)


jesus-vc:
@pallavi03 , hello. We're consumers of the HIPAA conformance pack. Do you mind sharing how this S3_BUCKET_LEVEL_PUBLIC_ACCESS_PROHIBITED rule is redundant? What other rules cover this protection? Thank you!

